PR #4, opened 2022-01-09
Co-authored-by: zainmfjavaid <zainmfjavaid@users.noreply.github.com>

Ariya3:


src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,45 @@
...
+import com.revrobotics.CANSparkMaxLowLevel;
+
+/** Add your docs here. */
+public class IntakeSubsystem extends SubsystemBase {
+  // Put methods for controlling this subsystem
+  // here. Call these from Commands.
+  private CANSparkMax m_motorController1 = new CANSparkMax(Constants.kMotorOneID, CANSparkMaxLowLevel.MotorType.kBrushless);
+  private CANSparkMax m_motorController2 = new CANSparkMax(Constants.kMotorTwoID, CANSparkMaxLowLevel.MotorType.kBrushless);
+  
+  public IntakeSubsystem(boolean reverse, boolean on){
```
azvanderpas: Turning on the motors shouldn't be done in the constructor. Make a new method for commands to use and leave configuration steps in the constructor.

src/main/java/frc/robot/Constants.java
```diff
@@ -12,4 +12,9 @@
  * <p>It is advised to statically import this class (or one of its inner classes) wherever the
  * constants are needed, to reduce verbosity.
  */
-public final class Constants {}
+public final class Constants {
+    public static final int kMotorOneID = 11;
```
sarango2158: Remember all of the subsystems will use the Constants file. How can we name the motors that belong to intake so they're easy to identify?

src/main/java/frc/robot/subsystems/HopperSubsystem.java
```diff
@@ -0,0 +1,34 @@
...
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.CANSparkMaxLowLevel;
+
+/** Add your docs here. */
+public class HopperSubsystem extends SubsystemBase {
+  // Put methods for controlling this subsystem
+  // here. Call these from Commands.
+  private CANSparkMax m_motorController3 = new CANSparkMax(Constants.kMotorThreeID, CANSparkMaxLowLevel.MotorType.kBrushless);
+  
+  public HopperSubsystem(boolean on){
```
azvanderpas: Same comment as on intake. Control motor speed in a different method

src/main/java/frc/robot/Constants.java
azvanderpas: Maybe add a TODO comment to rename these constants and ID numbering once the mechanisms have been designed.

azvanderpas (changes requested):
Good start. Don't forget to create a command class and hook them up to buttons as well.

src/main/java/frc/robot/subsystems/HopperSubsystem.java
sarango2158: This is your constructor, it is called just once when the HopperSubsystem is created (using "new"). It is used to assign values to any of the object's data and potentially set an initial state. If you want a way to control whether the Hopper is running at a given time, you should create a new method for that purpose.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
sarango2158: Same comment as for Hopper -- you should create a method to start, stop, or change the direction.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,55 @@
...
+  private CANSparkMax m_motorController2 = new CANSparkMax(Constants.kIntakeMotorTwoID, CANSparkMaxLowLevel.MotorType.kBrushless);
+  
+
+  public void IntakeController(){
+  
+  }
+  
+  public void IntakeSwitch(boolean on){
+    if (on){
+      m_motorController1.set(Constants.kIntakeMotorSpeed);
```
azvanderpas: Don't forget about the [follow function](https://codedocs.revrobotics.com/java/com/revrobotics/cansparkmax#follow(com.revrobotics.CANSparkMax)). This can be used to make one CANSparkMax always do the same thing as another. If this code becomes more complex it can save you a lot of copy/pasting

src/main/java/frc/robot/Robot.java
```diff
@@ -16,8 +19,9 @@
  */
 public class Robot extends TimedRobot {
   private Command m_autonomousCommand;
-
   private RobotContainer m_robotContainer;
+  private IntakeSubsystem m_intakeSubsystem;
```
azvanderpas, line 22: It would be better to instantiate the subsystems in RobotContainer because you'll want to use them with your Command classes and button bindings

zainmfjavaid (approved):

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -34,15 +44,20 @@ public RobotContainer() {
    * edu.wpi.first.wpilibj.Joystick} or {@link XboxController}), and then passing it to a {@link
    * edu.wpi.first.wpilibj2.command.button.JoystickButton}.
    */
-  private void configureButtonBindings() {}
+  private void configureButtonBindings() {
+    // TODO: figure out if we want to use whenPressed or WhileHeld
+    mButtons[Constants.kLeftBumperButton].whenPressed(mIntakeForwardCommand);
```
azvanderpas: Each index of mButtons (or at least the ones you want to use) needs to get initialized prior to this. Currently this will cause a runtime error.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,44 @@
...
+    if (on){
+      m_intakeMotorController1.set(Constants.kIntakeMotorSpeed);
+    } else {
+      m_intakeMotorController1.set(0);
+    }
+  }
+
+  public void ForwardIntake(){
+    m_intakeMotorController1.setInverted(false);
+    m_intakeMotorController2.setInverted(false);
```
azvanderpas: m_intakeMotorController2 references here and in ReverseIntake should be removed since its already setup to follow m_intakeMotorController1

src/main/java/frc/robot/subsystems/HopperSubsystem.java
```diff
@@ -0,0 +1,35 @@
...
+import frc.robot.Constants;
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.CANSparkMaxLowLevel;
+
+/** Add your docs here. */
+public class HopperSubsystem extends SubsystemBase {
+  // Put methods for controlling this subsystem
+  // here. Call these from Commands.
+  private CANSparkMax m_hopperController = new CANSparkMax(Constants.kHopperMotorThreeID, CANSparkMaxLowLevel.MotorType.kBrushless);
+  private CANSparkMax m_shoooterMotorController = new CANSparkMax(Constants.kShooterMotorID, CANSparkMaxLowLevel.MotorType.kBrushless);
```
azvanderpas: The shooter motor controller should only be instantiated by the Shooter subsystem. If you want to get the speed of the shooter motor, then a function that returns that must be provided by the shooter subsystem class. I'll try to cover this better at today's meeting.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -34,15 +44,20 @@ public RobotContainer() {
    * edu.wpi.first.wpilibj.Joystick} or {@link XboxController}), and then passing it to a {@link
    * edu.wpi.first.wpilibj2.command.button.JoystickButton}.
    */
-  private void configureButtonBindings() {}
+  private void configureButtonBindings() {
+    // TODO: figure out if we want to use whenPressed or WhileHeld
```
azvanderpas: The way your command classes are written, whileHeld makes more sense. Take a look these wpilib docs to get a better understanding:
[Binding Commands To Triggers](https://docs.wpilib.org/en/stable/docs/software/commandbased/binding-commands-to-triggers.html)
[The Scheduler Run Sequence](https://docs.wpilib.org/en/stable/docs/software/commandbased/command-scheduler.html#the-scheduler-run-sequence)

src/main/java/frc/robot/commands/IntakeReverseCommand.java
```diff
@@ -0,0 +1,40 @@
...
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {
+    m_intakeSubsystem.IntakeSwitch(false);
+  }
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    return false;
```
azvanderpas, line 38: Note that if you want to use whenPressed, isFinished will need to return true in some case (or some other action causes the Command to be canceled) in order for the command to end. As written, this command will start but never call the end function. See my other comment for wpilib docs to review.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,44 @@
...
+
+//import edu.wpi.first.wpilibj2.*;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Constants;
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.CANSparkMaxLowLevel;
+
+
+/** Add your docs here. */
+public class IntakeSubsystem extends SubsystemBase {
```
azvanderpas, line 15: This class is missing a constructor

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,44 @@
...
+  // here. Call these from Commands.
+  private CANSparkMax m_intakeMotorController1 = new CANSparkMax(Constants.kIntakeMotorOneID, CANSparkMaxLowLevel.MotorType.kBrushless);
+  private CANSparkMax m_intakeMotorController2 = new CANSparkMax(Constants.kIntakeMotorTwoID, CANSparkMaxLowLevel.MotorType.kBrushless);
+
+  public void IntakeController(){
+
+  }
+  
+  public void IntakeSwitch(boolean on){
+    m_intakeMotorController2.follow(m_intakeMotorController1, false);
```
azvanderpas, line 24: Move to constructor method

azvanderpas (changes requested):

azvanderpas (approved):

Issue #11, opened 2022-01-14
aim mode in ShooterSubsystem


caseyjbrotherton:
Using enum's in ShooterSubsystem for aimMode would allow for an easier to understand abstraction.
https://github.com/AusTINCANsProgrammingTeam/2022RobotCode/blob/d5579f37169a2eb71b6de95731a66fed51a0eca2/src/main/java/frc/robot/subsystems/ShooterSubsystem.java#L13

Reference:
https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Enum.html

If you have an enum for the aimMode, then you can use symbols instead of numbers in other code places.

Also of note, you can calculate the length of the enum instead of using a number.  Some thoughts here:

https://stackoverflow.com/questions/1741708/enum-values-length-vs-private-field

caseyjbrotherton:
Here might be an even cleaner way to handle increments of enums:  

https://stackoverflow.com/questions/17664445/is-there-an-increment-operator-for-java-enum

PR #25, opened 2022-01-19
Shooter prime

### LINKED ISSUE #22: Split motor control in CDS to operate independently
The CDS will consist of:

- One horizontal roller just after the intake system
- Two upright wheels funneling the cargo into a straight line (1-2 motors)
- One motor/wheel right before the shooter controlling when we decide to shoot
- (Maybe) an ejector system

Currently the CDS subsystem is all straight pass-through. We should set it up for independent control so that the shooter+CDS can work together. **Or, alternatively, put that motor into the shooter subsystem.**

loipiop224:
Merge Shooter Prime into main


src/main/java/frc/robot/subsystems/HopperSubsystem.java
```diff
@@ -9,10 +9,12 @@
 import com.revrobotics.CANSparkMax;
 import com.revrobotics.CANSparkMaxLowLevel;
 
-/** Add your docs here. */
+/** Add your docs here.
+ * Im sorry if this code is unreadable but deal with it
```
caseyjbrotherton: Let's remove this comment.  It's not helpful.

src/main/java/frc/robot/Constants.java
```diff
@@ -34,28 +35,59 @@
...
+    public static final double kLLAngle = 30.0; // Angle that the limelight is mounted at
+
+    public static final double kGoalHeight = 8.8; // Height of the goal in ft from the carpet
+    // Preset aim constants
+    public static final int kLOWRPM = 0; // RPM that the LOW aimMode winds to
+    public static final double kLOWAngle = 0.0; // Angle that the LOW aimMode adjusts to
+    public static final int kLAUNCHRPM = 0; // RPM that the LAUNCH aimMode winds to
+    public static final double kLAUNCHAngle = 0.0; // Angle that the LAUNCH aimMode adjusts to
+    public static final int kTARMACRPM = 0; // RPM that the TARMAC aimMode winds to
+    public static final double kTARMACAngle = 0.0; // Angle that the TARMAC aimMode adjusts to
```
caseyjbrotherton: Made a note in github issue: https://github.com/AusTINCANsProgrammingTeam/2022RobotCode/issues/11
about using an enum for tracking the different Aim Modes: Low/Launch/Tarmac/Auto

It makes sense to consider adding the RPM and Angle into the same enum

It would consolidate the constants, and allow you to directly return the angle, or RPM for a specific aim mode.

( The Planet example in the link here shows more:  https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html )

src/main/java/frc/robot/Constants.java
```diff
@@ -34,28 +35,59 @@
...
+
+    // Shooter Domain
+    public static final int KShooterID = 1; // ID of the shooter
+    public static final int KHoodID = 2; // ID of the hood;
+    public static final double KShooterHeight = 3;
+    public static final double KHighHeight = 8.8;
```
caseyjbrotherton: There is a kGoalHeight on line 71.  How is kGoalHeight different than KHighHeight?  

If it is a matter of aiming over the Hub, let's make it a bit clearer by adding a comment.  If they should be the same, consider combining them.

src/main/java/frc/robot/commands/ShooterModeCycleUp.java
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) FIRST and other WPILib contributors.
+// Open Source Software; you can modify and/or share it under the terms of
+// the WPILib BSD license file in the root directory of this project.
+
+package frc.robot.commands;
+
+import frc.robot.subsystems.ShooterSubsystem;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+
```
caseyjbrotherton: Let's add a short comment here on what the ShooterModeCycleUp command is doing.

If I understood, it changes the "type" of aiming that you will be doing with the shooter...Between several different positions on the field, or an automatic aiming.

Let's make it clearer with some comments

src/main/java/frc/robot/commands/ShooterPrimary.java
```diff
@@ -0,0 +1,44 @@
...
+
+import frc.robot.subsystems.ShooterSubsystem;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+
+public class ShooterPrime extends CommandBase {
+  private ShooterSubsystem m_ShooterSubsystem;
+
+  /** Creates a new ShooterPrimary. */
+  public ShooterPrime(ShooterSubsystem shooterSubsystem) {
```
caseyjbrotherton: ? This is a duplicate command as below.  

Seems we want to remove iti?

src/main/java/frc/robot/Constants.java
```diff
@@ -34,28 +35,59 @@
...
+    public static final double KLowHeight = 5.8;
+    public static final double KAirboneTime = 1.2;
+    public static final double KGravity = 32.0;
+    public static final double KGearRatioIn = 34;
+    public static final double KGearRatioOut = 14;
+    public static final double KGearDiametter = 4; // Gear diametter in inches
+    public static final double KBallFlywheelratio = 2;
+    public static final int KShooterCargoID = 6;
+    public static final double kCargoRotation = 3;
```
caseyjbrotherton: It's a small thing, but having consistent naming will help others navigate your code.  

Please change these to lowercase 'k' names.

src/main/java/frc/robot/Constants.java
```diff
@@ -34,28 +35,59 @@
...
+    public static final int KShooterID = 1; // ID of the shooter
+    public static final int KHoodID = 2; // ID of the hood;
+    public static final double KShooterHeight = 3;
+    public static final double KHighHeight = 8.8;
+    public static final double KLowHeight = 5.8;
+    public static final double KAirboneTime = 1.2;
+    public static final double KGravity = 32.0;
+    public static final double KGearRatioIn = 34;
+    public static final double KGearRatioOut = 14;
+    public static final double KGearDiametter = 4; // Gear diametter in inches
```
caseyjbrotherton: diameter has one t.  Please change to make it easier to search for the constant.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,170 @@
...
+     * return Fangle, velcocity1, velcocity2
+     */
+    x = x + 1;
+    double Fangle = Math.toDegrees(Math.atan((y - y0 + 1 / 2 * g * (Math.pow(t, 2))) / x));
+    double Velocity1 = Math.abs(x / (Math.cos(Math.toRadians(Fangle)) * t));
+    double Velocity2 = Math.abs((y - y0 + (g / 2.0) * (Math.pow(t, 2))) / (Math.sin(Math.toRadians(Fangle)) * t));
+    double[] VandA = new double[2];
+    VandA[0] = UnitConversion(Velocity1, Constants.KGearDiametter);
+    VandA[1] = Fangle;
+    return VandA;
```
caseyjbrotherton: Would it be better to return an object here?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  public double getDistance() {
+    // Uses Limelight to find distance to High Goal
+    return (Constants.kGoalHeight - Constants.kLLHeight) / Math.tan(getTY() + Constants.kLLAngle); // Return distance in
+                                                                                                   // feet
+  }
+
+  public double[] ProjectilePrediction(double y0, double x0, double y, double x, double g, double t) {
+    /**
```
caseyjbrotherton: Would like to see a brief commend explaining what is expected to be passed into this method.

I feel like G does not need to be passed in.  It can be just referenced directly.  

If there is a reference to the physics that you used, it would be good to place it in the comments as well.


Would like to talk through this section.  Velocity 2 is not used.  t is a parameter that maps to a constant.

Velocity 1 looks like the x based velocity.  velocity 2 looks like the y based velocity.  
Velocity 1 is returned, but that is not the required velocity of the Cargo, because it is only the X component? 

Not sure, need to think through it, but we should talk further about it.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,170 @@
...
+     * Fangle =math.degrees(math.atan((y-y0+1/2*g*(t**2))/x))
+     * velcocity1 = abs(x/(math.cos(math.radians(Fangle))*t))
+     * velcocity2 = abs((y-y0+(g/2.0)*(t**2))/(math.sin(math.radians(Fangle))*t))
+     * return Fangle, velcocity1, velcocity2
+     */
+    x = x + 1;
```
caseyjbrotherton: Is this to prevent non-zero x?  Would it be better to just return an error if there is a zero x?

caseyjbrotherton (changes requested):

src/main/java/frc/robot/commands/ShooterModeCycleDown.java
```diff
@@ -0,0 +1,42 @@
...
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {
+  }
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    return false;
```
azvanderpas: Using the whenPressed trigger means you have to determine end conditions or it'll keep running. In this case you probably just want to change the return value to true since you want to initialize and then end.  

You can save time and code by converting this into an [InstantCommand](https://docs.wpilib.org/en/stable/docs/software/commandbased/convenience-features.html#instantcommand)

src/main/java/frc/robot/subsystems/HopperSubsystem.java
```diff
@@ -9,10 +9,12 @@
...
+ */
 public class HopperSubsystem extends SubsystemBase {
   // Put methods for controlling this subsystem
-  // here. Call these from Commands.
+  // here. Call these PIDfrom Commands.
```
azvanderpas: Search and replace typo?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,170 @@
...
+    KHoodController.setReference(a, CANSparkMax.ControlType.kPosition);
+
+    // Adjusts Hood using PID control to passed angle a
+  }
+
+  public void windFlywheel(int rpm) {
+
+    double kGearRationRPM = rpm * (Constants.KGearRatioIn / Constants.KGearRatioOut);
+    KShooterController.setReference(kGearRationRPM, CANSparkMax.ControlType.kVelocity);
+    while (true) {
```
azvanderpas: Instead of doing this 'blocking' check, create a separate method to check for ready once each 'execute' phase of your ShooterPrime command (or in 'isFinished'). Then have another method to run the CargoController motor in the 'end' phase. This prevents your command from blocking other scheduled commands while you wait for the shooter to be ready.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,170 @@
...
+
+      double KFlywheelspeed = KShooterEncoder.getVelocity();
+      if (KFlywheelspeed == rpm) {
```
azvanderpas: When checking RPM, don't expect us to hit our target with exact accuracy. Check for a range rather then one value.

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
azvanderpas: We had worked through this together somewhat. I thought this was more for proof of concept though. Velocity1 and velocity2 should always be the same since we are solving the projectile x and y equations for the velocity magnitude

x = x<sub>0</sub> + v * cos(angle)      and    y = y<sub>0</sub> + v * sin(angle) - gt<sup>2</sup>/2

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: For our current purposes, I feel like a creating classes with objects is not yet necessary, and that it may be better to just use an array output at the moment.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: This is an offset applied to x so that we are shooting to the center of our goal, rather than the rims, where the vision tape is.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: Added a range of +/- 15 from our RPM to shoot for.

src/main/java/frc/robot/Constants.java
```diff
@@ -62,4 +57,32 @@
...
+    // Preset aim constants
+    public static final int kLOWRPM = 0; // RPM that the LOW aimMode winds to
+    public static final double kLOWAngle = 0.0; // Angle that the LOW aimMode adjusts to
+    public static final int kLAUNCHRPM = 0; // RPM that the LAUNCH aimMode winds to
+    public static final double kLAUNCHAngle = 0.0; // Angle that the LAUNCH aimMode adjusts to
+    public static final int kTARMACRPM = 0; // RPM that the TARMAC aimMode winds to
+    public static final double kTARMACAngle = 0.0; // Angle that the TARMAC aimMode adjusts to
+
+    // Shooter Constants
+    public static final int kShooterID = 1; // ID of the shooter
```
sarango2158: Conflict with motor ID in drive base (kDriveRightMiddle)

src/main/java/frc/robot/Constants.java
```diff
@@ -62,4 +57,32 @@
...
+    public static final int kHoodID = 2; // ID of the hood;
```
sarango2158: Conflict with motor ID in drive base (kDriveRightRear)

src/main/java/frc/robot/Constants.java
```diff
@@ -62,4 +57,32 @@
...
+    public static final double kLowHeight = 5.0 + 7.75/12.0; // Height of the low goal in ft from the carpet
+    public static final double kAirboneTime = 1.2;
+    public static final double kGravity = 32;
+    public static final double kLLHeight = 1.5; // Height of the limelight in ft from the carpet
+    public static final double kLLAngle = 30.0; // Angle that the limelight is mounted at
+    public static final double kGearRatioIn = 34;
+    public static final double kGearRatioOut = 14;
+    public static final double kGearDiameter = 4; // Gear diametter in inches
+    public static final double kBallFlywheelratio = 2;
+    public static final int kShooterCargoID = 6;
```
sarango2158: Conflict with motor ID in drive base (kDriveRightFront)

sarango2158 (changes requested):

src/main/java/frc/robot/Constants.java
sarango2158: Ignore this, drive base team is removing ID 6 now that we only have 4 drive motors

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: Open an issue for this.

src/main/java/frc/robot/commands/ShooterPrime.java
```diff
@@ -34,11 +34,16 @@ public void execute() {
...
+
   }
 
   // Returns true when the command should end.
   @Override
   public boolean isFinished() {
+    if(m_ShooterSubsystem.wheelReady()){
+      m_ShooterSubsystem.shoot();
+      return true;
+    }
```
azvanderpas: Nothing functionally wrong here. but isFinished() really shouldn't do anything other than return whether the command is done. Put shoot() in the end() method, and check interrupted there to make sure the flywheel is actually ready and the command didn't get canceled for some other reason.

azvanderpas:
It looks like a merge at some point reverted some of the changes to resolve my previous comments.

src/main/java/frc/robot/commands/ShooterModeCycleDown.java
zbrotherton: Converting these to InstantCommands.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -4,24 +4,91 @@
...
   }
 
   public void adjustHood(double a) {
-    //Adjusts Hood using PID control to passed angle a
+    KHoodController.setReference(a, CANSparkMax.ControlType.kPosition);
+
+    // Adjusts Hood using PID control to passed angle a
   }
 
   public void windFlywheel(int rpm) {
```
sarango2158: Should we pass this as a double instead, so you don't need to cast it?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -4,24 +4,91 @@
...
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.CANSparkMaxLowLevel;
+
+import org.opencv.core.Mat;
+import com.revrobotics.RelativeEncoder;
+import frc.robot.common.hardware.MotorController;
 
 public class ShooterSubsystem extends SubsystemBase {
-  private int aimMode; //0 is LOW, 1 is AUTO, 2 is LAUNCH, 3 is TARMAC
+  private int aimMode; // 0 is LOW, 1 is AUTO, 2 is LAUNCH, 3 is TARMAC
```
sarango2158: 4 is TEST

sarango2158 (approved):

src/main/java/frc/robot/commands/ShooterPrime.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {
+    if (!interrupted){
+      //TODO: uncomment when cargoMotor exists 
+      //m_ShooterSubsystem.shoot()
+    }
+    m_ShooterSubsystem.windFlywheel(0); //TODO: Look at this again, we don't want our flywheel to instantly switch off in the end
```
caseyjbrotherton, line 42: Does it make sense to keep the flywheel coasting?  Then you could turn the flywheel off, and let it coast until you need to shoot again.

zbrotherton: It would, but can I do that? Maybe I can just add voltage control to shooterSubsystem if rpm == 0, so it doesn't actively try to get to 0? Is there a better way to do that?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -30,29 +97,83 @@ public void setAimMode(int m) {
...
-    return NetworkTableInstance.getDefault().getTable("limelight").getEntry("ty").getDouble(0);
+    // Gets TY, the vertical angle of the target from the limelight
+    return Math.toDegrees(NetworkTableInstance.getDefault().getTable("limelight").getEntry("ty").getDouble(0));
   }
+
   public double getDistance() {
-    //Uses Limelight to find distance to High Goal
-    return (Constants.kGoalHeight - Constants.kLLHeight) / Math.tan(getTY() + Constants.kLLAngle); //Return distance in feet
+    // Uses Limelight to find distance to High Goal
+    return (Constants.kHighHeight - Constants.kLLHeight) / Math.tan(getTY() + Constants.kLLAngle); // Return distance in
```
caseyjbrotherton: math.tan requires radians.  https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#tan-double-

https://github.com/AusTINCANsProgrammingTeam/common-lib/blob/main/src/main/java/hardware/Limelight.java#L59

zbrotherton: Does that mean we need to redo our calculations, or should I just convert to degrees after .tan is run

src/main/java/frc/robot/commands/ShooterPrime.java
caseyjbrotherton: I would recommend opening an issue to track it, and go ahead and close this comment.  I can poke into REV's example code, and post links in the issue.  

I mean, you could change it to m_ShooterSubsystem.flywheelOff() or something, and then the subsystem code can do the right thing. 

But that can be a later action.

zbrotherton: I just don't know how to let the wheel coast, it slows down pretty quickly. I will add a simple voltage control, so that the motor isn't slowing things down, and is simply just getting no power.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
caseyjbrotherton: Link to the common code has an example:
            Math.tan(Math.toRadians(mLimelightMountingAngle + mLimelightTable.getEntry("ty").getDouble(0.0))); 

So...For your code:
    return (Constants.kHighHeight - Constants.kLLHeight) / Math.tan(Math.toRadians(getTY() + Constants.kLLAngle)); // Return distance in

caseyjbrotherton (approved):

src/main/java/frc/robot/commands/ShooterPrime.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    if(m_ShooterSubsystem.wheelReady()){
+      return true;
+    }
+    return false;
+  }
```
azvanderpas, line 52: You can just do:
`return m_ShootSubsystem.wheelReady()`
The if-else block isn't necessary

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -4,24 +4,91 @@
...
+    Thread.sleep(200);
+    cargo_motorController.setSpeed(0.0);*/
+  }
+
+  public boolean wheelReady(){
+    double flywheelSpeed = KShooterEncoder.getVelocity();
+    if (flywheelSpeed > currentRPM - 15 && flywheelSpeed < currentRPM + 15) {
+      return true;
+    }
+    return false;
```
azvanderpas, line 91: You can just return the expression here and not use the if-else block

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -30,29 +97,83 @@ public void setAimMode(int m) {
...
+    }
   }
 
   public void cycleAimModeDown() {
     aimMode--;
-    if(aimMode < 0){aimMode = 3;}
+    if (aimMode < 0) {
+      aimMode = 3;
+    }
   }
```
azvanderpas, line 110: If you cycle out of aimMode=4, you can't get back to it

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -30,29 +97,83 @@ public void setAimMode(int m) {
...
+    double[] returnArray = new double[2];
+    returnArray[0] = UnitConversion(velocity, Constants.kGearDiameter); //Converts FPS to RPM
+    returnArray[1] = hoodAngle;
+    return returnArray;
+
+  }
+
+  public double UnitConversion(double KBallSpeed, double GearDiameter) {
+    // Convert from FPS of the ball into RPM
+    return (((KBallSpeed * 12) / Constants.kGearDiameter) * Constants.kBallFlywheelratio) * 2;
```
azvanderpas, line 136: Documentation on this equation would be useful to include for future reference.

azvanderpas (approved):
Approved with comments that can be addressed later

PR #36, opened 2022-01-22
Intake

### LINKED ISSUE #26: Remove naming prefix letters
[Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s5.1-identifier-names) says don't use naming prefixes

I think I agree they are not very helpful, so we should try to remove them and stop using them.

Ariya3:


src/main/java/frc/robot/Constants.java
```diff
@@ -14,60 +14,56 @@
...
-    public static final int kCDSMotorThreeID = 13;
-    public static final int kCDSMotorFourID = 14;
-    public static final int kCDSMotorFiveID = 15;
-
-    public static final double kCDSWheelSpeed = 0.25;
-    public static final double kCDSBeltSpeed = 0.25;
+    public static final int CDSBeltID = 100;
+    public static final int CDSMotorID = 101;
+    public static final double CDSBeltSpeed = 0.70;
+    public static final int ballSensorChannel = 0;
```
sarango2158: Will need multiple channels, one for each Digital Input

src/main/java/frc/robot/Constants.java
```diff
@@ -14,60 +14,56 @@
...
     //Joystick Constants
-    public static final int kPortNumber = 0;
-    public static final int kJoystickButtonNumberOne = 1;
-    public static final int kJoystickButtonNumberTwo = 2;
-    public static final int kBButton = 2; // Button for hopper
-    public static final int kXButton = 3; // Button for reverse hopper
-    public static final int kRightBumperButton = 5; // Button for intake 
-    public static final int kLeftBumperButton = 6;  // Button to reverse intake
+    public static final int portNumber = 0;
+    public static final int joysticbuttonNumberOne = 1;
```
sarango2158: This isn't used, let's get rid of it

src/main/java/frc/robot/Constants.java
```diff
@@ -14,60 +14,56 @@
...
+    public static final int joysticbuttonNumberTwo = 2;
```
sarango2158: This also isn't used, let's get rid of it

src/main/java/frc/robot/Constants.java
```diff
@@ -14,60 +14,56 @@
...
 
     //CDS Constants
-    public static final int kCDSMotorThreeIndex = 0;
-    public static final int kCDSMotorThreeID = 13;
-    public static final int kCDSMotorFourID = 14;
-    public static final int kCDSMotorFiveID = 15;
-
-    public static final double kCDSWheelSpeed = 0.25;
-    public static final double kCDSBeltSpeed = 0.25;
+    public static final int CDSBeltID = 100;
```
sarango2158: What are these used for on the robot? Do we need this?

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -8,51 +8,86 @@
...
-    CDSWheelControllerOne =  new MotorController("Main CDS Wheel Controller", Constants.kCDSMotorFourID);
-    CDSWheelControllerTwo = new MotorController("Follows CDS Wheel Controller", Constants.kCDSMotorFiveID);
-    
-    CDSWheelControllerTwo.getSparkMax().follow(CDSWheelControllerOne.getSparkMax());
+    CDSBeltController = new MotorController("CDS Motor", Constants.CDSBeltID);
+    CDSMotorController2 = new MotorController("CDS Motor 2", Constants.CDSMotorID);
+   
+    intitalBallSensor = new DigitalInput(Constants.ballSensorChannel);
+    middleBallSensor = new DigitalInput(Constants.ballSensorChannel);
+    finalBallSensor = new DigitalInput(Constants.ballSensorChannel);
```
azvanderpas: Are these all meant to be on the same DIO?

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -8,51 +8,86 @@
...
+      CDSBeltController.getSparkMax().set(Constants.CDSBeltSpeed);
+      SmartDashboard.putNumber("CDS Belt Speed", Constants.CDSBeltSpeed);
     } else {
       CDSBeltController.getSparkMax().set(0.0);
-      CDSWheelControllerOne.getSparkMax().set(0.0);
       SmartDashboard.putNumber("CDS Motor Speed", 0);
     }
   }
 
   public void ForwardCDS() {
```
sarango2158: ForwardCDS() and ReverseCDS() are the same command and just toggle the direction of the speed. You should either do this with one Toggle() method, so it is clear that it is state-independent, or actually have ForwardCDS() and ReverseCDS() confirm that the motor is moving in the direction intended. 

For example, if you call ForwardCDS() twice in a row, it will go in opposite directions each time.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -8,51 +8,86 @@
...
     SmartDashboard.putString("CDS Belt Direction", "Reverse");
-    SmartDashboard.putString("CDS Wheel Direction", "Reverse");
   }
 
+  public boolean getDirection() {
+    // true = inverted, false = forward
+    if (CDSBeltController.getSparkMax().get() > 0) {
+      return false;
+    } else {
+    return CDSBeltController.getSparkMax().getInverted();
```
sarango2158: We're not using inversion on the motors anymore, will this method give you the information you are looking for?

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -8,51 +8,86 @@
...
+  public void indexBall() {
+    boolean initialSensorStatus = getInitialSensorStatus();
+    boolean finalSensorStatus = getFinalSensorStatus();
+    if (!initialSensorStatus)  {
+      if (!this.getDirection()){
+        ballCount++;
+      } else {
+        ballCount--;
+      }
+      if (ballCount > 2) {
```
sarango2158: I would lean towards moving the logic to take action on having too many cargo into a different method, instead of making it part of indexBall().

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -8,51 +8,86 @@
...
-    SmartDashboard.putString("CDS Wheel Direction", "Forward");
   }
 
   public void ReverseCDS() {
-    CDSBeltController.getSparkMax().setInverted(true);
-    CDSWheelControllerOne.getSparkMax().setInverted(true);
+    CDSBeltController.getSparkMax().set(-Constants.CDSBeltSpeed);
     SmartDashboard.putString("CDS Belt Direction", "Reverse");
-    SmartDashboard.putString("CDS Wheel Direction", "Reverse");
   }
```
azvanderpas: These now make CDSSwitch redundant except when you want to turn the motors off.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -8,51 +8,86 @@
...
+  }
+  public boolean getMiddleSensorStatus(){
+    return middleBallSensor.get();
+  }
+  public boolean getFinalSensorStatus(){
+    return finalBallSensor.get();
+  }
+  public void indexBall() {
+    boolean initialSensorStatus = getInitialSensorStatus();
+    boolean finalSensorStatus = getFinalSensorStatus();
```
azvanderpas: Seems like if you used an array you could count these in a simpler way.

sarango2158 (dismissed):

src/main/java/frc/robot/Constants.java
sarango2158: CDSBeltID and CDSMotorID ^

zainmfjavaid: We don't need them right now, but when the CDS is finished there should be a separate belt that this constant gets used for

caseyjbrotherton (approved):

PR #46, opened 2022-01-29
intake merge request

### LINKED ISSUE #58: CDS Subsystem Duplicate
CDS Subsystem was making two of the same motor ID

Ariya3:


src/main/java/frc/robot/Robot.java
```diff
@@ -24,20 +25,21 @@
...
-  private Command m_autonomousCommand;
-  private RobotContainer m_robotContainer;
-  private TabContainer m_tabContainer;
-
-  // This function is run when the robot is first started up and should be used for any
-  // initialization code.
+  private Command autonomousCommand;
+  private RobotContainer robotContainer;
+  private TabContainer tabContainer;
+  private IntakeSubsystem ballController;
```
AndrewIverson: Intake subsystem should be exist and be created inside of robot container, not inside Robot.java.

src/main/java/frc/robot/Robot.java
```diff
@@ -87,14 +89,16 @@ public void teleopInit() {
...
+    if (autonomousCommand != null) {
+      autonomousCommand.cancel();
     }
   }
 
   // This function is called periodically during operator control. */
   @Override
-  public void teleopPeriodic() {}
+  public void teleopPeriodic() {
+    ballController.expelBalls();
```
AndrewIverson: This should be a command that is schedule and is always running, not called from Robot.java

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -49,22 +51,23 @@
 
 
   private final Joystick driverJoystick = new Joystick(Constants.portNumber);
-  private JoystickButton[] mButtons = new JoystickButton[11];
+  private JoystickButton[] buttons = new JoystickButton[11];
```
AndrewIverson, line 54: m_buttons to match our coding standards. All private variables would have m_name

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -49,22 +51,23 @@
...
 
   // subsystems
   private final DriveBaseSubsystem driveBaseSubsystem = new DriveBaseSubsystem(driverJoystick);
   private final CDSSubsystem CDSSubsystem = new CDSSubsystem();
-  private final IntakeSubsystem IntakeSubsystem = new IntakeSubsystem(); 
-  private final ShooterSubsystem ShooterSubsystem = new ShooterSubsystem();
+  private final IntakeSubsystem intakeSubsystem = new IntakeSubsystem();
```
AndrewIverson, line 60: Rename to have m_intake.... same for shooterSubsystem

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -8,28 +8,27 @@
...
 import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
 import frc.robot.common.hardware.MotorController;
+//import edu.wpi.first.wpilibj.DigitalInput;
 
 /** Add your docs here. */
 public class CDSSubsystem extends SubsystemBase {
   // Put methods for controlling this subsystem
   // here. Call these from Commands.
   private MotorController CDSBeltController;
+  private MotorController CDSAlignmentWheelOne;
```
AndrewIverson: Private variables should be named m_name.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -19,35 +20,84 @@
...
+    return true;
     }
   }
-  
-  public void ForwardIntake(){
-    intakeMotorControllerOne.getSparkMax().setInverted(false);
-    SmartDashboard.putString("Intake Motor Direction", "Forward");
+
+  public boolean[] getBeamBreakStatus() {
+    boolean initialStatus = initialBallSensor.get();
```
AndrewIverson: Suggestion: rename intialStatus to frontStatus, initial implies starting status and that is not really accurate here.

src/main/java/frc/robot/subsystems/Tabs/TabContainer.java
```diff
@@ -4,7 +4,7 @@
 
 package frc.robot.subsystems.Tabs;
 
-import frc.robot.subsystems.Tabs.TabDriveBase;
+//import frc.robot.subsystems.Tabs.TabDriveBase;
```
AndrewIverson, line 7: I don't think this should stay commented out.

AndrewIverson (changes requested):
Overall looking good

src/main/java/frc/robot/subsystems/CDSSubsystem.java
azvanderpas: Google Java Style guide doesn't use prefixes:

[5.1 Rules common to all identifiers](https://google.github.io/styleguide/javaguide.html#s5.1-identifier-names)
"In Google Style, special prefixes or suffixes are not used. For example, these names are not Google Style: name_, mName, s_name and kName."

I noted this in #26 which is why they are being removed here.

src/main/java/frc/robot/RobotContainer.java
azvanderpas: See reply below about prefixes, but I would avoid making the variable name the same as the class name. At least make it lowerCamelCase to match the style guide

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -19,35 +20,84 @@
...
+    boolean[] beamBreakStatuses = this.getBeamBreakStatus();
+    if (!beamBreakStatuses[0]) {
+      if (this.getDirection()){
+        ballCount--;
+      } else {
+        ballCount++;
+      }
+    }
+    return ballCount;
   }
```
azvanderpas: A few things here:
1. It seems like the ball count should just count the beam breaks rather than try to detect which way the motion is. At least add some comments to explain the intent. 
2. If the intake is not moving, and we have 1 ball stored, ballCount will keep incrementing and cause us to spit it out. Set ballCount to 0 before checking all the statuses.
3. I think we talked about only using the first beam break to testing right now, but add a TODO comment to update it later.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -19,35 +20,84 @@
...
 
-  
+  public void expelBalls() {
+    if (ballCount > 2) {
+      this.toggleIntake(true);
+      while (this.getBallCount() > 2) {
+      }
+      this.toggleIntake(false);
```
azvanderpas: End state should probably be to stop the motor rather than switch its direction. Consider renaming toggleIntake to something like runIntake. 

Since it seems like the intent here is to potentially override a button press, so maybe using the while loop is a good way to do that. But currently this implements a blocking wait which may cause issues with other subsystems if it takes too long. I'm not sure how well the scheduler handles that.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
AndrewIverson: Interesting. The biggest reason I see to do it is it makes it impossible(maybe unlikely?) to accidentally use the wrong scope of a variable. Like a parameter that is named the same as a class's private variable. 

I've resolved my comments about the prefixes.

src/main/java/frc/robot/subsystems/Tabs/TabContainer.java
azvanderpas: VSCode marks this as unused/unnecessary. I think its because TabDriveBase is in the same folder/package?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -89,7 +92,7 @@ public RobotContainer() {
       e.printStackTrace();
     }
 
-    driveBaseSubsystem.setDefaultCommand(mDriveBaseTeleopCommand);
+    driveBaseSubsystem.setDefaultCommand(driveBaseTeleopCommand);
```
azvanderpas, line 97: Instead of putting it in Robot.java, I think you want to make a new InstantCommand for your expelBalls method and make it the default command for the intake subsystem.

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
AndrewIverson: The while loop could possibly trigger motor watch dogs which will kill all motor outputs. Looping like this should be avoided.

src/main/java/frc/robot/RobotContainer.java
sarango2158: We may need _slightly_ more logic in the command to know when to stop expelling, Zain and Ariya are debating making a full command instead

src/main/java/frc/robot/commands/BeamBreakCommand.java
```diff
@@ -5,22 +5,23 @@
...
-import frc.robot.subsystems.CDSSubsystem;
+import frc.robot.subsystems.IntakeSubsystem;
 
-public class CDSAlignmentWheelCommand extends CommandBase {
-  /** Creates a new CDSAlignmentWheelCommand. */
-  private final CDSSubsystem mCDSAlignmentWheel;
+public class BeamBreakCommand extends CommandBase {
+  /** Creates a new BeamBreakCommand. */
+  public final IntakeSubsystem mIntakeSubsystem; 
+  IntakeSubsystem intakeSubsystem;
```
azvanderpas, line 13: Delete this duplicate

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -95,9 +96,99 @@ public int getBallCount() {
   public void expelBalls() {
     if (ballCount > 2) {
```
azvanderpas, line 98: I think you want to call getBallCount() here instead of using the ballCount variable. Currently getBallCount() never gets run anywhere so ballCount will aways be 0.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -95,9 +96,99 @@ public int getBallCount() {
...
+
+
+
+
+
+
+
+
+
+
```
azvanderpas: delete extra whitespace

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -19,35 +20,86 @@
...
+  private DigitalInput middleBallSensor;
+  private DigitalInput finalBallSensor;
+
+  private int ballCount = 0;
 
   public IntakeSubsystem() {
-    intakeMotorControllerOne = new MotorController("Intake Motor One", Constants.intakeMotorOneID);
-    intakeMotorControllerTwo = new MotorController("Intake Motor Two", Constants.intakeMotorTwoID); 
+    intakeMotorControllerOne = new MotorController("Intake Motor One", Constants.intakeMotorOneID, 40);
+    intakeMotorControllerTwo = new MotorController("Intake Motor Two", Constants.intakeMotorTwoID, 40);
```
sarango2158: We don't need this either, intake now only has one motor for the 4 rollers.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
sarango2158: Skipping this for now in the PR

sarango2158 (approved):

PR #48, opened 2022-01-29
Motor controller updates

### LINKED ISSUE #13: getPID() could return null
https://github.com/AusTINCANsProgrammingTeam/2022RobotCode/blob/d5579f37169a2eb71b6de95731a66fed51a0eca2/src/main/java/frc/robot/common/hardware/MotorController.java#L56

If the default constructor is used, mPIDController isn't initialized. getPID() should handle this in some way

### LINKED ISSUE #33: If PID controller is unintilizied we should throw an error or something to make it apparent
https://github.com/AusTINCANsProgrammingTeam/2022RobotCode/blob/ab60cc198bbdfc5ff1ab8952528e955be3bc01b6/src/main/java/frc/robot/common/hardware/MotorController.java#L57

Comes from the issue Zane had, it wasn't apparent that the getPID() was returning something that wasn't a working PID controller.

### LINKED ISSUE #34: Add default smartCurrentLimit setting   to CANSparkMax in MotorController class

### LINKED ISSUE #37: Add method to set a motor to be brake or coast within motorController class

sarango2158:
Fixes issues 33, 34, and 37 with the MotorController class by adding a method (setIdleMode), putting a default current limit in the default constructor of the class, and adding some error handling for potentially uninitialized objects such as mPIDController and mEncoder.

### DIFF
--- a/src/main/java/frc/robot/common/hardware/MotorController.java
+++ b/src/main/java/frc/robot/common/hardware/MotorController.java
@@ -23,16 +23,20 @@ public class MotorController {
     public MotorController(String name, int deviceID) {
         mName = name;
         mSparkMax = new CANSparkMax(deviceID, MotorType.kBrushless);
+        mSparkMax.restoreFactoryDefaults();
+
+        // Create default values for Spark Max motor controller
         mSparkMax.setSmartCurrentLimit(40); // default current limit is 40A
+        mSparkMax.setIdleMode(CANSparkMax.IdleMode.kCoast); // default mode is Coast
 
         mEncoder = mSparkMax.getEncoder();
-        mSparkMax.restoreFactoryDefaults();
     }
 
 
     public MotorController(String name, int deviceID, int smartCurrentLimit, boolean... enablePid) {
         this(name, deviceID);                               // intializes CANSparkMax and Encoder
         mSparkMax.setSmartCurrentLimit(smartCurrentLimit);  // set smartCurrentLimit
+        mSparkMax.setIdleMode(CANSparkMax.IdleMode.kCoast); // default mode is Coast
 
         // If enablePid has any number of booleans greater than 0 we are enabling pid
         if (enablePid.length > 0)
@@ -49,7 +53,7 @@ public class MotorController {
     public CANSparkMax getSparkMax() {
         // Check first that mSparkMax has been instantiated
         if(mSparkMax == null) {
-            throw new NullPointerException("Spark MAX motor has not been instantiated.");
+            throw new NullPointerException("Spark MAX motor for "+this.mName +" has not been instantiated.");
         }
         return mSparkMax;
     }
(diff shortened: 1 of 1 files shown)


ModBoyex (approved):

zbrotherton (approved):

tonyn10 (approved):

src/main/java/frc/robot/common/hardware/MotorController.java
```diff
@@ -23,6 +23,7 @@
     public MotorController(String name, int deviceID) {
         mName = name;
         mSparkMax = new CANSparkMax(deviceID, MotorType.kBrushless);
+        mSparkMax.setSmartCurrentLimit(40); // default current limit is 40A
 
         mEncoder = mSparkMax.getEncoder();
         mSparkMax.restoreFactoryDefaults();
```
azvanderpas: Its unclear from the Rev Robotics docs, but this might override the smart current limit setting. Probably should move up to right after mSparkMax is instantiated.

src/main/java/frc/robot/common/hardware/MotorController.java
```diff
@@ -110,4 +132,9 @@ public void updateSmartDashboard() {
         }
     }
 
+    // Mode can be coast or brake
+    public void setIdleMode(CANSparkMax.IdleMode mode) {
+        mSparkMax.setIdleMode(mode);
```
azvanderpas, line 141: Should we set an explicit default in the constructor?

src/main/java/frc/robot/common/hardware/MotorController.java
```diff
@@ -46,14 +47,26 @@ public MotorController(String name, int deviceID, int smartCurrentLimit, boolean
...
+            throw new NullPointerException("Encoder has not been instantiated.");
+        }
         return mEncoder;
     }
 
-    public SparkMaxPIDController getPID() {
+    public SparkMaxPIDController getPID() { 
+        // Check first that mPIDController has been instantiated
+        if(mPIDController == null) {
+            throw new NullPointerException("PID Controller has not been instantiated.");
```
azvanderpas: Could we add the mName to the error message? Though if mSparkMax doesn't exist, mName probably doesn't either.

azvanderpas (approved):

src/main/java/frc/robot/common/hardware/MotorController.java
sarango2158: The default for Spark Max in general is Coast, but I agree it'd be easier to see that explicitly called out

src/main/java/frc/robot/common/hardware/MotorController.java
sarango2158: I think we can get the name of the MotorController, we wouldn't get this far if the MotorController itself didn't exist. Added!

PR #52, opened 2022-02-02
Ll align

zbrotherton:


src/main/java/frc/robot/commands/LimelightAlign.java
```diff
@@ -0,0 +1,45 @@
...
+
+import frc.robot.subsystems.LimelightSubsystem;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+
+public class LimelightAlign extends CommandBase {
+  private LimelightSubsystem m_LimelightSubsystem;
+
+  /** Creates a new ShooterPrimary. */
+  public LimelightAlign(LimelightSubsystem LimelightSubsystem) {
+    addRequirements(LimelightSubsystem);
```
azvanderpas: Add the drivebase subsystem as a requirement as well.

src/main/java/frc/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,75 @@
...
+  private CANSparkMax m_leftMotor;
+  private CANSparkMax m_rightMotor;
+  private boolean isFinished;
+
+  public LimelightSubsystem() {
+    m_PidController = new PIDController(6e-3, 0, 0);
+    m_PidController.setTolerance(2.0);
+    m_DriveBaseSubsystem = RobotContainer.getDriveBase();
+    m_leftMotor = m_DriveBaseSubsystem.getLeftMotor();
+    m_rightMotor = m_DriveBaseSubsystem.getRightMotor();
```
azvanderpas: Instead of having to deal with the motors directly here, it would be better to add a setSpeed method to Drivebasesubsystem and call that. 

Then if there's changes to the motor configuration it doesn't mess up your code

src/main/java/frc/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,75 @@
...
+    double calculation = MathUtil.clamp(m_PidController.calculate(getTX(), 0.0), -1.0, 1.0);
+      // Uses TX and our setpoint (which will always be 0.0) to return the next calculation
+    if (m_PidController.atSetpoint()){ // If our robot is aligned within the tolerance, return 0.0 to end command
+      SmartDashboard.putNumber("Finished?", 1);
+      System.out.println("done");
+      isFinished = true;
+      return 0.0;
+    }
+    else{
+      return Math.signum(calculation) * Math.max(Math.abs(calculation),0.05);
```
azvanderpas, line 47: Probably worth a comment to document why this is necessary.

src/main/java/frc/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,75 @@
...
+    m_rightMotor = m_DriveBaseSubsystem.getRightMotor();
+    isFinished = false;
+  }
+
+  public double getTX() {
+    // Gets TX, the horizontal angle of the target from the limelight
+    return NetworkTableInstance.getDefault().getTable("limelight").getEntry("tx").getDouble(0);
+  }
+
+  public double calculatePID(){
```
azvanderpas: This looks like it should a private method since its just for helping setMotors

src/main/java/frc/robot/commands/LimelightAlign.java
```diff
@@ -0,0 +1,45 @@
...
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    m_LimelightSubsystem.setMotors();
+  }
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {
+    m_LimelightSubsystem.setMotors();
```
azvanderpas: Does setMotors need to be called again here? Seems like the motors should have already stopped. 

Should also explicitly stop the motors here in case the command is interrupted/canceled before it's finished.

src/main/java/frc/robot/commands/LimelightAlign.java
caseyjbrotherton: Talking with @Majorferguson21 it's unsure to me if you need to add drivebase as a requirement here.

 @azvanderpas thoughts?

caseyjbrotherton (approved):

src/main/java/frc/robot/subsystems/LimelightSubsystem.java
```diff
@@ -0,0 +1,75 @@
...
+    m_PidController.setTolerance(2.0);
+    m_DriveBaseSubsystem = RobotContainer.getDriveBase();
+    m_leftMotor = m_DriveBaseSubsystem.getLeftMotor();
+    m_rightMotor = m_DriveBaseSubsystem.getRightMotor();
+    isFinished = false;
+  }
+
+  public double getTX() {
+    // Gets TX, the horizontal angle of the target from the limelight
+    return NetworkTableInstance.getDefault().getTable("limelight").getEntry("tx").getDouble(0);
```
azvanderpas, line 36: Once case to be aware of is what this will return if there is no vision target on camera and what that will cause the align command to do. 

Doesn't need to be addressed yet in this PR.

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/LimelightSubsystem.java
zbrotherton: Good point, I was going to try giving a full value to the motors until tx is found, basing which way it turns off of the gyro, maybe.

src/main/java/frc/robot/commands/LimelightAlign.java
```diff
@@ -0,0 +1,47 @@
...
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    m_LimelightSubsystem.stopMotors();
+  }
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {
+    m_LimelightSubsystem.setMotors();
```
azvanderpas: I think you flipped stopMotors and setMotors

src/main/java/frc/robot/commands/LimelightAlign.java
```diff
@@ -0,0 +1,47 @@
...
+
+import frc.robot.subsystems.DriveBaseSubsystem;
+import frc.robot.subsystems.LimelightSubsystem;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+
+public class LimelightAlign extends CommandBase {
+  private LimelightSubsystem m_LimelightSubsystem;
+
+  /** Creates a new ShooterPrimary. */
+  public LimelightAlign(LimelightSubsystem LimelightSubsystem, DriveBaseSubsystem DriveBaseSubsystem) {
```
azvanderpas: make the argument names lowerCamelCase so they are distinct from the class names

azvanderpas (approved):
I think ideally all the inter-subsystem control would all happen within a command (with subsystem requirements for each) rather then having one subsystem directly access another.  But I'm fine having that as a future improvement.

PR #55, opened 2022-02-05
Issue #40 fix

### LINKED ISSUE #40: Indicate to operator when robot is ready to shoot
When the robot has at least one ball, is positioned, and is spun up for a shot, there should be an indicator to the human operator that they can press shoot. 

Options:
1. An indicator on the driver dashboard software
2. CAD created a physical sign we can use that will be lit by LEDs. Can connect that up to USB and turn the LEDs ON when the robot is ready to shoot.

zbrotherton:
#40 should be fixed by this, also finishes the full automatic shoot process.

### DIFF
--- a/src/main/java/frc/robot/commands/ShooterPrime.java
+++ b/src/main/java/frc/robot/commands/ShooterPrime.java
@@ -4,17 +4,21 @@
 
 package frc.robot.commands;
 
+import frc.robot.subsystems.LimelightSubsystem;
 import frc.robot.subsystems.ShooterSubsystem;
 import edu.wpi.first.wpilibj2.command.CommandBase;
 
 public class ShooterPrime extends CommandBase {
   private ShooterSubsystem m_ShooterSubsystem;
+  private LimelightSubsystem m_LimelightSubsystem;
   private int i;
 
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/commands/ShooterPrime.java
```diff
@@ -4,17 +4,22 @@
...
+  private LimelightSubsystem m_LimelightSubsystem;
   private int i;
 
   /** Creates a new ShooterPrimary. */
-  public ShooterPrime(ShooterSubsystem shooterSubsystem) {
+  public ShooterPrime(ShooterSubsystem shooterSubsystem, LimelightSubsystem limelightSubsystem) {
     addRequirements(shooterSubsystem);
     m_ShooterSubsystem = shooterSubsystem;
+    m_LimelightSubsystem = limelightSubsystem;
+    SmartDashboard.putBoolean("wheelReady", false);
```
caseyjbrotherton, line 22: Would be better if we considered adding these to a different tab in shuffleboard.

We can determine the full UI later, but at least we will start with them on a new tab.

caseyjbrotherton (approved):

PR #59, opened 2022-02-05
Subsys Disable 

azvanderpas:
Shuffleboard controlled subsystem disabling ultimately was too buggy. Settled on using JSON file to enable/disable subsystems

### DIFF
--- a/src/main/deploy/SubsystemEnables.json
+++ b/src/main/deploy/SubsystemEnables.json
@@ -1,0 +1,8 @@
+[
+    {
+        "CDSSubsystem"       : "true",
+        "DriveBaseSubsystem" : "true",
+        "IntakeSubsystem"    : "true",
+        "ShooterSubsystem"   : "true"
+    }
+]
--- a/src/main/deploy/SubsystemEnables.json
+++ b/src/main/deploy/SubsystemEnables.json
@@ -3,6 +3,7 @@
         "CDSSubsystem"       : "true",
         "DriveBaseSubsystem" : "true",
         "IntakeSubsystem"    : "true",
-        "ShooterSubsystem"   : "true"
+        "ShooterSubsystem"   : "true",
+        "LimeligthSubsystem" : "true"
     }
 ]
--- a/src/main/deploy/SubsystemEnables.json
+++ b/src/main/deploy/SubsystemEnables.json
@@ -1,9 +1,0 @@
-[
-    {
-        "CDSSubsystem"       : "true",
-        "DriveBaseSubsystem" : "true",
-        "IntakeSubsystem"    : "true",
-        "ShooterSubsystem"   : "true",
-        "LimeligthSubsystem" : "true"
-    }
-]
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -76,26 +76,8 @@ public class RobotContainer {
   // private int trajectoryIndex = 0;
   private Trajectory trajectory;
 
-  private JSONObject subSysEnables;
-
   // The container for the robot. Contains subsystems, OI devices, and commands.
   public RobotContainer() {
-    try {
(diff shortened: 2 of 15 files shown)


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -52,34 +55,105 @@
...
+      JSONArray jsonArray = (JSONArray) JSONValue.parse(jsonString);
+      subSysEnables = (JSONObject) jsonArray.get(0);
+       
+      //for (Object k : subSysEnables.keySet()) {
+      //   System.out.println((String) k + " " + subSysEnables.get((String) k));
+      //}
+
+
+    } catch (IOException e) {
+      //TODO catch/fail gracefully
```
caseyjbrotherton: Trying to think through this.  

What is the graceful failure?  What is the best case here if there is a failure?

I think that the best case is for everything to be turned on.  It would be annoying in test for this type of failure, but safer in competition.

Thinking through that, and seeing the typo above, makes me wonder if we might flip this paradigm around.

Change SystemEnable.json -> SystemDisable.json, any subsystem listed there is _not_ enabled.

src/main/deploy/SubsystemEnables.json
```diff
@@ -0,0 +1,9 @@
+[
+    {
+        "CDSSubsystem"       : "true",
+        "DriveBaseSubsystem" : "true",
+        "IntakeSubsystem"    : "true",
+        "ShooterSubsystem"   : "true",
+        "LimeligthSubsystem" : "true"
```
caseyjbrotherton: typo

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -52,34 +55,105 @@
...
       mButtons[i] = new JoystickButton(driverJoystick, i);
     }
 
+    // subsystems
+    for (Object k : subSysEnables.keySet()) {
+      if (((String)subSysEnables.get((String) k)).equals("true")) {
+
+        //System.out.println((String) k + " " + subSysEnables.get((String) k));
+        switch ((String) k) {
+          case "DriveBaseSubsystem":
```
caseyjbrotherton, line 90: enum for these?  
Then we can iterate through the enum and check what is set/not set.

src/main/java/frc/robot/RobotContainer.java
caseyjbrotherton: Or I mentioned below, you could use an enum for the subsystems, iterate through those, and if you find a missing subsystem, assume it's true.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -52,34 +55,105 @@
...
     debugTab = Shuffleboard.getTab("debug");
+
     // Configure the button bindings
     for (int i = 1; i < mButtons.length; i++) {
       mButtons[i] = new JoystickButton(driverJoystick, i);
     }
 
+    // subsystems
+    for (Object k : subSysEnables.keySet()) {
+      if (((String)subSysEnables.get((String) k)).equals("true")) {
```
caseyjbrotherton: "true".equals((String)subSysEnables.get((String) k))

if you change the iteration at all, then using the constant prevents NPE's

caseyjbrotherton (changes requested):
My main thought is on the graceful failure.  Notes are in the comments, but probably out of order. :)

If you want the failure case to be every subsystem on, then you can approach it in two different ways:

1) Change the system to disable subsystems, and if the JSON is not found, or failed to read, then you don't disable any subsystems.
2) Iterate through an enum of subsystems instead of what is in the JSON.  If you don't find a subsystem, then you by default turn it on.

caseyjbrotherton (approved):
looks good!

sarango2158 (approved):

azvanderpas:
Fixed merge conflicts, but PR didn't update. Going to resubmit

PR #65, opened 2022-02-08
Intake

### LINKED ISSUE #18: Inter-subsystem communication
Need a way for subsystems to get data from each other (e.g. Hopper needs to know when Shooter is ready)

There are probably other ways to do this, but one way is to create public functions in RobotContainer that return the subsystems it instantiates. Then a subsystem can get a reference to another with (e.g.) RobotContainer.getShooterSubsystem().

### LINKED ISSUE #57: Constants Conflicts
In the Constants file, there were two fives conflicting with each other.

zbrotherton:


src/main/java/frc/robot/commands/ShooterPrime.java
```diff
@@ -50,12 +54,13 @@ public void end(boolean interrupted) {
...
     if(m_ShooterSubsystem.wheelReady()){
       SmartDashboard.putBoolean("wheelReady", true);
-      if(i > 0 || m_LimelightSubsystem.calculatePID() == 0.0){
-        m_ShooterSubsystem.runCargo(true);
+      //if(i > 0 || m_LimelightSubsystem.calculatePID() == 0.0){
+        m_ShooterSubsystem.runCargo(true,true);
+        m_CDSSubsystem.CDSBeltWheelControllerToggle(false);
         i++;
-        if(i==10){ //Expected to add a 200ms delay
+        if(i==100){ //Expected to add a 200ms delay
```
caseyjbrotherton: No longer 200ms with 100.  Should we add a constant here?

src/main/java/frc/robot/commands/ShooterPrime.java
```diff
@@ -50,12 +54,13 @@ public void end(boolean interrupted) {
   public boolean isFinished() {
     if(m_ShooterSubsystem.wheelReady()){
       SmartDashboard.putBoolean("wheelReady", true);
-      if(i > 0 || m_LimelightSubsystem.calculatePID() == 0.0){
-        m_ShooterSubsystem.runCargo(true);
+      //if(i > 0 || m_LimelightSubsystem.calculatePID() == 0.0){
```
caseyjbrotherton, line 57: Disabled because there is no limelight...Should we consider instead returning 0.0 from calculatePID?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -26,29 +23,23 @@
...
   private MotorController cargo_motorController;
-  private SparkMaxPIDController kCargoController;
-  private RelativeEncoder kCargoEncoder;
-  private NetworkTableEntry sbShooterRPM;
   private double currentRPM;
 
   public ShooterSubsystem() {
-    sbShooterRPM = RobotContainer.debugTab.add("shooterRPM", 0).getEntry();
-
+    SmartDashboard.putNumber("RPMChange", 3000.0);
```
caseyjbrotherton: Think this was removed, and not committed?  Didn't work, maybe?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -26,29 +23,23 @@
...
     aimMode = 4;
     cargo_motorController = new MotorController("Shooter Cargo", Constants.shooterCargoID);
-    kCargoController = cargo_motorController.getPID();
-    kCargoEncoder = cargo_motorController.getEncoder();
-
     shooter_motorController = new MotorController("Shooter", Constants.shooterID, 40, true);
     KShooterController = shooter_motorController.getPID();
     KShooterEncoder = shooter_motorController.getEncoder();
-    KShooterController.setP(6e-4);
+    KShooterController.setP(6e-5);
```
caseyjbrotherton: Check your local, because I seem to remember 3e something being tests.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -26,29 +23,23 @@
...
     KShooterController.setI(1e-6);
     KShooterController.setD(0.0);
-
+    KShooterController.setOutputRange(0, 1);
+    /*
```
caseyjbrotherton, line 40: Let's add TODO comments where we are marking these sections as removed for testing.  
Reason being is that they will then show up in vscode as something to change.

caseyjbrotherton (changes requested):

src/main/java/frc/robot/commands/ShooterPrime.java
zbrotherton: I don't know how we could use constants here, I'll make sure to keep the comment up to date until I can rework a new system.

src/main/java/frc/robot/commands/ShooterPrime.java
zbrotherton: It would be cleaner, but a harder change to remember, I would have no idea what was going wrong.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: It was a test attempt by someone that never worked but was never removed, I think

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -113,16 +111,15 @@ private void configureButtonBindings() {
     buttons[Constants.rightBumperButton].whileHeld(intakeReverseCommand);
     
     // CDS
-    buttons[Constants.YButton].whileHeld(CDSForwardCommand);
-    buttons[Constants.BButton].whileHeld(CDSReverseCommand);
+    buttons[Constants.leftBumperButton].whileHeld(CDSForwardCommand);
+    buttons[Constants.rightBumperButton].whileHeld(CDSReverseCommand);
```
sarango2158: in conflict with cycling aim mode down (rightBumperButton = downbutton in Constants)

src/main/java/frc/robot/commands/CDSForwardCommand.java
```diff
@@ -4,23 +4,31 @@
...
+    addRequirements(shooterSubsystem);
+    mShooterSubsystem = shooterSubsystem;
     mCDSSubsystem = CDSSubsystem;
   }
 
   // Called when the command is initially scheduled.
   @Override
   public void initialize() {
     mCDSSubsystem.CDSBeltWheelControllerToggle(false);
+    mShooterSubsystem.runCargo(true, false);
```
sarango2158, line 30: Fine for now while we're testing, but may want a nicer handoff between belt and cargo stopper, such that we would avoid overrunning the stopper with the belt

caseyjbrotherton (approved):
Talked about changes with Zane, asked him to check with Siana or Adam as well.

src/main/java/frc/robot/commands/CDSReverseCommand.java
```diff
@@ -5,22 +5,27 @@
...
+    addRequirements(shooterSubsystem);
+    mShooterSubsystem = shooterSubsystem;
     mCDSSubsystem = CDSSubsystem;
   }
 
   // Called when the command is initially scheduled.
   @Override
   public void initialize() {
     mCDSSubsystem.CDSBeltWheelControllerToggle(true);
+    mShooterSubsystem.runCargo(true, false);
```
sarango2158, line 28: The overall command here reverses the CDS. Do we want the cargo stopper to do anything?

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -28,21 +31,22 @@ public CDSSubsystem() {
 
   public void CDSBeltWheelControllerToggle(boolean reverse) {
     if (reverse) {
-      CDSWheelControllerOne.getSparkMax().set(-Constants.CDSWheelControllerSpeed);
+      CDSWheelControllerOne.getSparkMax().set(Constants.CDSWheelControllerSpeed);
       SmartDashboard.putString("CDS Wheel Controller Direction", "Reverse");
       SmartDashboard.putNumber("CDS Wheel Controller Speed", -Constants.CDSWheelControllerSpeed);
       
       CDSBeltController.getSparkMax().set(-Constants.CDSBeltSpeed);
+      CDSBeltController.getSparkMax().setIdleMode(IdleMode.kBrake);
```
sarango2158: MotorController class should be able to set this directly now in its methods (both speed and Idle mode)

sarango2158 (changes requested):

src/main/java/frc/robot/commands/CDSReverseCommand.java
zbrotherton: We do, actually. This is because there is a possibility the driver forced the ball too far into the shooter, and needs to reverse it.

src/main/java/frc/robot/commands/CDSForwardCommand.java
zbrotherton: True, I hate the method call myself (if that's what you mean by handoff) but I can't get boolean... to work in attributes.
If you meant a second command or a delay, I think sequencing that might be a bit too hard, because the stopper is only running for the unpredictable chance a second ball pushes the top ball through the stopper.

sarango2158 (approved):

PR #67, opened 2022-02-10
Climb base

### LINKED ISSUE #42: Create basic climb subsystem
- Motor to pull spring depresser
- Motor on winch to climb
- Camera pointed vertically to align to bar with

ModBoyex:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -54,22 +56,26 @@
...
   private static CDSSubsystem CDSSubsystem;
   private static IntakeSubsystem intakeSubsystem; 
   private static ShooterSubsystem shooterSubsystem;
   private static LimelightSubsystem limelightSubsystem;
 
   // commands
-  private final DriveBaseTeleopCommand driveBaseTeleopCommand = new DriveBaseTeleopCommand(driveBaseSubsystem);
-  private IntakeForwardCommand intakeForwardCommand = new IntakeForwardCommand(intakeSubsystem);
-  private IntakeReverseCommand intakeReverseCommand = new IntakeReverseCommand(intakeSubsystem);
+  private DriveBaseTeleopCommand driveBaseTeleopCommand;
```
sarango2158, line 67: I made these changes so those commands would be initialized inside of the modular switch case.

src/main/java/frc/robot/commands/ClimbDOWNCammand.java
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) FIRST and other WPILib contributors.
```
sarango2158: Spelled "Command" wrong

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -54,22 +56,26 @@
...
+  private IntakeForwardCommand intakeForwardCommand;
+  private IntakeReverseCommand intakeReverseCommand;
 
    // private BeamBreakCommand beamBreakCommand = new BeamBreakCommand(intakeSubsystem);
-  private ShooterPrime shooterPrime = new ShooterPrime(shooterSubsystem,limelightSubsystem,CDSSubsystem);
-  private CDSForwardCommand CDSForwardCommand = new CDSForwardCommand(CDSSubsystem,shooterSubsystem);
-  private CDSReverseCommand CDSReverseCommand = new CDSReverseCommand(CDSSubsystem,shooterSubsystem);
-  private LimelightAlign limelightAlign = new LimelightAlign(limelightSubsystem,driveBaseSubsystem);
+  private ClimbUPCamand climbUPCommand;
+  private ClimbDOWNCammand climbDOWNCammand;
```
sarango2158: Spelled "Command" wrong

src/main/java/frc/robot/commands/ClimbUPCamand.java
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) FIRST and other WPILib contributors.
```
sarango2158: Spelled "Command" wrong

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,66 @@
...
+  
+
+  // TODO: maybe add servos
+
+  public ClimbSubsystem() {
+    //One is left, two is right
+    m_climbMotorControllerOne = new MotorController("Climb Motor One", Constants.kClimbMotorOneIndex);
+
+    m_climbMotorControllerTwo = new MotorController("Climb Motor Two", Constants.kClimbMotorTwoIndex);
+    m_climbMotorControllerTwo.setInverted(true);
```
sarango2158, line 31: Why do we set inverted here if the motors are taking the same action in the same direction?

sarango2158 (changes requested):

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public void enableClimb(boolean on, boolean up){
+    if (on) {
+      if (up) {
+        m_climbMotorControllerOne.getPID().setReference(25, CANSparkMax.ControlType.kPosition);
+        SmartDashboard.putNumber("Climb Hight", m_climbMotorControllerOne.getEncoder().getPosition());
+      } else {
+        m_climbMotorControllerOne.getPID().setReference(-25, CANSparkMax.ControlType.kPosition);
+        SmartDashboard.putNumber("Climb Hight", m_climbMotorControllerOne.getEncoder().getPosition());
```
azvanderpas: Move the SmartDashboard calls to periodic() or somewhere else where they will get updated.

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,66 @@
...
+    m_climbMotorControllerTwo.setInverted(true);
+    m_climbMotorControllerTwo.setFollow(m_climbMotorControllerOne);
+
+    m_limitSwitch = new DigitalInput(Constants.kLimitSwitchChannel);
+
+    //Example camera system, unsure if it goes here or not
+    CameraServer.startAutomaticCapture();
+    //Not used at the moment, maby in the futue
+    CvSink cvSink = CameraServer.getVideo();
+    CvSource outputStream = CameraServer.putVideo("Blur", 640, 480);
```
azvanderpas: Remove this for now since it isn't used

src/main/java/frc/robot/Constants.java
```diff
@@ -142,4 +168,7 @@ public final Boolean isEnabled() {
     public static final double ballFlywheelratio = 2;
     public static final int shooterCargoID = 5;
     public static final double cargoRotation = 3;
-}
\ No newline at end of file
+
+    public static final int kClimbMotorOneIndex = 1;
+    public static final int kClimbMotorTwoIndex = 2;
```
azvanderpas: You are using these as motorIDs, so naming them Index isn't correct

src/main/java/frc/robot/Constants.java
```diff
@@ -116,6 +118,30 @@ public final Boolean isEnabled() {
...
+    //Climb Constants
+    //Check top of constants
+    //TODO: Fill in placeholder values
+
+
+    //CDS Constants
+    //Check top of constants
+    public static final double kCDSWheelSpeed = 0.25;
+    public static final double kCDSBeltSpeed = 0.25;
+
```
azvanderpas: Don't these already exists somewhere?

src/main/java/frc/robot/Constants.java
```diff
@@ -116,6 +118,30 @@ public final Boolean isEnabled() {
     public static final int downPOV = 4; // Button for shooter mode
     public static final int upPOV = 5; // Button for shooter mode
     public static final int AButton = 2; //Button to align Limelight
+    public static final int LTrigger = 9; //Button For Climb
+    public static final int RTrigger = 8; //Button For Climb
```
azvanderpas: As we tested, these don't map to the triggers

azvanderpas (changes requested):

src/main/java/frc/robot/Constants.java
```diff
@@ -4,6 +4,7 @@
 
 package frc.robot;
 import edu.wpi.first.math.kinematics.DifferentialDriveKinematics;
+import frc.robot.subsystems.ClimbSubsystem;
```
caseyjbrotherton: Dont think this is needed?

src/main/java/frc/robot/commands/ClimbUPCamand.java
```diff
@@ -0,0 +1,41 @@
...
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {
+    m_subsystem.enableClimb(false, false);
+  }
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    return false;
```
caseyjbrotherton: Should we be checking that the enableClimb setreference is finished here? 
 ( I assume that would require another method in the subsystem. )
 
 Not seeing a sparkmax code example for using the position

caseyjbrotherton (changes requested):

src/main/java/frc/robot/commands/ClimbUPCamand.java
ModBoyex: I'm not sure I'm pretty sure that that code was there when I made the Command

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
ModBoyex: The motors for the Climer are going to face different directions meaning that one of the motors needs to be inverted

src/main/java/frc/robot/Constants.java
```diff
@@ -19,10 +19,11 @@
     public enum Subsystems {
         // Change booleans to disable a subsystem in RobotContainer
         DriveBaseSubsystem(false),
-        CDSSubsystem      (true),
-        IntakeSubsystem   (true),
-        ShooterSubsystem  (true),
-        LimelightSubsystem(false);
+        CDSSubsystem      (false),
```
sarango2158: Re-enable these subsystems before completing your PR

src/main/java/frc/robot/Robot.java
```diff
@@ -4,9 +4,9 @@
 
 package frc.robot;
 
-
 import edu.wpi.first.wpilibj.TimedRobot;
 import edu.wpi.first.wpilibj2.command.Command;
+import frc.robot.subsystems.ClimbSubsystem;
```
sarango2158: We should not do anything with specific commands or subsystems in Robot.java. Those should be handled through the Command Scheduler by setting up commands in RobotContainer -- the way you have your commands set up in RobotContainer already should be enough. Please remove all references to ClimbSubsystem in this file.

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,60 @@
...
+      } else {
+        m_climbMotorControllerOne.getEncoder().setPosition(0);
+        m_climbMotorControllerOne.getPID().setReference(-25, CANSparkMax.ControlType.kPosition);
+      }
+    } else {
+      m_climbMotorControllerOne.setSpeed(0);
+    }
+  }
+
+  public void periodic() {
```
sarango2158, line 53: Instead of putting this in Robot.java, recommend combining this logic with enableClimb. Alternatively, if you want this to always be running instead of tied to the up/down commands, you could make this method into an InstantCommand and set it as the default for the subsystem

sarango2158 (changes requested):

sarango2158 (approved):

caseyjbrotherton (approved):

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
ModBoyex: Moved it to Up/Down command so that it updates everytime the command is run

azvanderpas (approved):

PR #80, opened 2022-02-12
Drive Base

### LINKED ISSUE #17: Autonomous drivebase requirements
Mostly just writing notes on what we need to do to get to running trajectories

1. Drive base needs encoders on all the wheels for drive characterization (ones built into the SparkMax/NEOs will not work)
2. Add Gyro to drivebase class
3. Run Robot Characterization project to get Ks, Kv, and Ka constants
4. Create Trajectories to use with PathWeaver
5. Implement methods required by RamseteCommand:
                      Supplier<Pose2d> pose,    (Gyro input)
                      Supplier<DifferentialDriveWheelSpeeds> wheelSpeeds,  (encoder input)
                      BiConsumer<Double,​Double> outputVolts  (special function to set drive voltages directly)

There is a simpler RamseteCommand constructor that only requires:
                      Supplier<Pose2d> pose, (Gyro)
                      BiConsumer<Double,​Double> outputMetersPerSecond, (wheel speed targets rather than volts, no PID)

The BiConsumer and Supplier parameters to RamseteCommand are 'lambdas', which is a way to pass a function as an parameter for the RamseteCommand class to use, in this case for both getting info about our drivebase (Supplier), and sending speed data to our drivebase function (BiConsumer).

6. Put it all together. Example project here: https://github.com/wpilibsuite/allwpilib/tree/main/wpilibjExamples/src/main/java/edu/wpi/first/wpilibj/examples/ramsetecommand

### LINKED ISSUE #62: Debug Tab Creation Needs To Be Moved
Robot Container, Line 48. 

The Debug Tab needs to be created here before Subsystems Call it.

### LINKED ISSUE #72: RobotContainer has constructors outside of modular subsystem
Possibly due to a bad merge, RobotContainer currently calls the constructors of all of the commands outside of the switch case for the modular subsystem. It also sets the default command for the DriveBase subsystem. This causes null references when we turn off subsystems in Constants because the subsystems are never created, but are being used for the related commands.

tonyn10:
Autonomous implemented. Motor controller constants updated. Drive should be smoother and rotation is correct.


Paths/Unnamed.path
```diff
@@ -0,0 +1,3 @@
+X,Y,Tangent X,Tangent Y,Fixed Theta,Reversed,Name
```
azvanderpas: Give this a name if you want to keep it in Git

simgui.json
```diff
@@ -2,8 +2,11 @@
   "NTProvider": {
```
azvanderpas: This is an autogenerated file. Lets not track it in Git

src/main/deploy/paths/Unnamed.wpilib.json
```diff
@@ -0,0 +1,932 @@
+[
```
azvanderpas: Name file if you want to keep it

src/main/java/frc/robot/commands/BeamBreakCommand.java
```diff
@@ -14,7 +14,9 @@
 
   public BeamBreakCommand(IntakeSubsystem intakeSubsystem) {
     // Use addRequirements() here to declare subsystem dependencies.
-    addRequirements(intakeSubsystem); 
+    if(intakeSubsystem != null) {
+      addRequirements(intakeSubsystem); 
+    }
```
azvanderpas: This check alone isn't helpful, because it'll just crash the first time it tries to use the intakesubsystem if it is null. 
Another PR got merged that broke the subsystem disables in RobotContainer. The shooter subsystem PR shows what should be there. 

Either remove these checks or add else { cancel() } to unschedule them if their subsystems don't exist. I'm not certain about the cancel method, so you'd have to test it if you go that route.

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
```diff
@@ -61,43 +67,97 @@ public DriveBaseSubsystem(Joystick joystick) {
...
  
   }
 
   // Normal Arcade Drive
   public void arcadeDrive() {
-    m_differentialDrive.arcadeDrive(m_driverJoystick.getRawAxis(Constants.leftJoystickY), 
-                                    m_driverJoystick.getRawAxis(Constants.rightJoystickX));
+    m_differentialDrive.arcadeDrive(-1 * m_driverJoystick.getRawAxis(Constants.leftJoystickY), 
+                                        m_driverJoystick.getRawAxis(Constants.rightJoystickX));
+    // x-axis of rotation joystick is inverted so multiply by -1
```
azvanderpas: Stale comment

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
```diff
@@ -61,43 +67,97 @@ public DriveBaseSubsystem(Joystick joystick) {
...
+
+    this.usingExternal = usingExternal; // gets key if using external or internal encoders
+
+    // // external encoders            
+    // m_extLeftEncoder = new Encoder(Constants.leftEncoderDIOone, Constants.leftEncoderDIOtwo, 
+    //                          false, Encoder.EncodingType.k2X); // TODO: confirm correct configuration for encoder
+    // m_extRightEncoder = new Encoder(Constants.rightEncoderDIOone, Constants.rightEncoderDIOtwo, 
+    //                          false, Encoder.EncodingType.k2X);
+
+    // m_extLeftEncoder.setReverseDirection(true);
```
azvanderpas: If usingExternal gets true, you're code with crash because the m_ext*Encoders aren't initialized.

azvanderpas (changes requested):

azvanderpas (approved):

PR #82, opened 2022-02-13
Add Spotless to build.gradle

### LINKED ISSUE #16: Setup code formatting enforcment
There are other options, but wpilibs has a doc for Spotless:
https://docs.wpilib.org/en/stable/docs/software/advanced-gradlerio/code-formatting.html

azvanderpas:


caseyjbrotherton (commented):
What is your plan for rollout?  
This will break a build if there is a format violation.
We will need to teach the team about opening the terminal to run ./gradlew.bat :spotlessApply

Should we consider: https://marketplace.visualstudio.com/items?itemName=richardwillis.vscode-spotless-gradle ?

I thought we had discussed this, and decided that static style checking would be better for pre-season.

caseyjbrotherton:
Talked with Adam, if it automatically applied, then it might be usable for the team.

https://stackoverflow.com/questions/40432291/gradle-spotless-task-not-firing-when-needed#40433262

Followed this pattern seemed to work both with "./gradlew build" and "Build Robot Code" from vscode.

( Changed the spaces before a package command in a java  file, and build set it back. )

We should bring this up in standup before merging so that the team is aware of the changes.

caseyjbrotherton (approved):

PR #83, opened 2022-02-13
Add Simulation run to CI build pipeline

azvanderpas:


src/main/java/frc/robot/Constants.java
```diff
@@ -157,7 +157,7 @@ public final Boolean isEnabled() {
     public static final double goalHeight = 8.8; //Height of the goal in ft from the carpet
     
     // Shooter Constants
-    public static final int shooterID = 4; // ID of the shooter
+    public static final int shooterID = 14; // ID of the shooter
```
azvanderpas: Conflicts with drivebase motor ids (kitbot).

.github/workflows/main.yml
```diff
@@ -32,3 +31,11 @@ jobs:
     # Runs a single command using the runners shell
     - name: Compile and run tests on robot code
       run: ./gradlew build
+
+    # Runs a simulation to spot check crashes
+    - name: Run simulation to check for runtime errors
+      run: timeout 1m ./gradlew simulateJava 2>&1 | awk -v rc=0 '/The robot program quit unexpectedly/ { rc=1 } 1; END {exit rc}'
```
caseyjbrotherton, line 37: If I run the command above without Robot errors, then the return code is 143. ( SIGTERM sent )

seems that awk is receiving the sigterm from timeout?

caseyjbrotherton (changes requested):

.github/workflows/main.yml
azvanderpas: It seems to work in GitHub's linux environment (CI pipeline is clean). Where are you running the command?

caseyjbrotherton: Running Linux extensions for windows.  There might be subtle difference in signaling.  will go ahead and close this if it works in github's linux

caseyjbrotherton (approved):

src/main/java/frc/robot/Constants.java
sarango2158: On MDF bot:
driveRightFront = 7
driveRightRear = 8
driveLeftFront = 1
driveLeftRear = 2
intakeMotorOneID = 3
CDSBeltID = 6
intakeWheelOneID = 12
intakeWheelTwoID = 11
shooterCargoID = 5
shooterID = 4

sarango2158 (approved):

PR #88, opened 2022-02-16
Shooter testing

### LINKED ISSUE #11: aim mode in ShooterSubsystem

### LINKED ISSUE #53: Store shooting parameters in a lookup table
Instead of calculating the different parameters of the shooter on the fly, store values for e.g. distance, flywheel RPM, hood angle in a lookup table to reference more quickly. Possibly add interpolation for in-between values.

### LINKED ISSUE #101: Make a new command that shoots for autonomous command.
Command for shooting now runs while button held.

Need a new command that runs the flywheel, and shoots automatically, and ends when the balls are shot.

This would be used for autonomous

@loipiop224 @Majorferguson21

zbrotherton:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -153,13 +150,15 @@ private void initCommands() {
     if (CDSSubsystem != null && shooterSubsystem != null) {
       CDSForwardCommand = new CDSForwardCommand(CDSSubsystem);
       CDSReverseCommand = new CDSReverseCommand(CDSSubsystem);
+      // CDSSubsystem.getAllianceColor();
+      CDSSubsystem.senseColor();
```
sarango2158: Was this intentional? You moved this method call from configureButtonBindings() to initCommands()

caseyjbrotherton (approved):
looks good

src/main/java/frc/robot/commands/ShooterHeld.java
```diff
@@ -0,0 +1,79 @@
...
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    m_ShooterSubsystem.prime();
+    if (m_ShooterSubsystem.wheelReady()) {
+      if (i > 0 || !LLEnabled || m_LimelightSubsystem.calculatePID() == 0.0) {
+        i++;
+        m_ShooterSubsystem.runCargo(true, true);
+        m_ShooterSubsystem.setCargoBoolean(true);
+        if (i >= 50) {// 1000 miliseconds delay
```
sarango2158: If the delay is here to make sure we've shot all of the cargo in the robot, we can potentially poll our CDS sensors to see if they're all clear and finish the command more precisely

src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -40,35 +42,36 @@ public void initialize() {
...
 
   // Returns true when the command should end.
   @Override
   public boolean isFinished() {
+    if (m_ShooterSubsystem.wheelReady()) {
+      if (i > 0 || !LLEnabled || m_LimelightSubsystem.calculatePID() == 0.0) {
+        i++;
+        m_ShooterSubsystem.runCargo(true, true);
+        m_ShooterSubsystem.setCargoBoolean(true);
+        if (i >= 50) {// 1000 miliseconds delay
```
sarango2158: Same comment as from ShooterHeld -- if this is to make sure we've shot all the cargo, polling the CDS sensors could be more accurate

sarango2158 (changes requested):

src/main/java/frc/robot/commands/ShooterHeld.java
```diff
@@ -0,0 +1,79 @@
...
+  private LimelightSubsystem m_LimelightSubsystem;
+  private CDSSubsystem m_CDSSubsystem;
+  private int i;
+  private boolean LLEnabled;
+
+  /** Creates a new ShooterHeld. */
+  public ShooterHeld(
+      ShooterSubsystem shooterSubsystem,
+      LimelightSubsystem limelightSubsystem,
+      CDSSubsystem cdsSubsystem,
```
azvanderpas, line 24: CDSSubsystem isn't used, should we removed it from the parameter list?

src/main/java/frc/robot/commands/ShooterHeld.java
```diff
@@ -0,0 +1,79 @@
...
+      boolean llEnabled) {
+    addRequirements(shooterSubsystem);
+    m_ShooterSubsystem = shooterSubsystem;
+    m_LimelightSubsystem = limelightSubsystem;
```
azvanderpas, line 31: Use llEnabled to determine whether to do addRequirements(limelightSubsystem)

src/main/java/frc/robot/commands/ShooterHeld.java
```diff
@@ -0,0 +1,79 @@
...
+  public void initialize() {
+    i = 0;
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    m_ShooterSubsystem.prime();
+    if (m_ShooterSubsystem.wheelReady()) {
+      if (i > 0 || !LLEnabled || m_LimelightSubsystem.calculatePID() == 0.0) {
```
azvanderpas, line 51: High level comments would be useful here to understand what the intent of this logic is.

src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -40,35 +42,36 @@ public void initialize() {
...
+          return true;
+        }
 
-    return false;
+      } else {
+        m_ShooterSubsystem.runCargo(false, false);
```
azvanderpas: Move the runCargo calls to the end() method. isFinished should just tell us whether the command is done and nothing else.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -153,13 +150,15 @@ private void initCommands() {
...
     }
     if (intakeSubsystem != null) {
       intakeForwardCommand = new IntakeForwardCommand(intakeSubsystem);
       intakeReverseCommand = new IntakeReverseCommand(intakeSubsystem);
     }
     if (shooterSubsystem != null && limelightSubsystem != null && CDSSubsystem != null) {
```
azvanderpas:
```suggestion
    if (shooterSubsystem != null && CDSSubsystem != null) {
       shooterHeld = new ShooterHeld(shooterSubsystem, limelightSubsystem, CDSSubsystem,(limelightSubsystem != null));
```

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -72,7 +76,7 @@
   private NetworkTableEntry ShooterReverted =
       shooterTab.add("Shooter Reverted", false).withPosition(1, 4).getEntry();
   private NetworkTableEntry DShootingMode =
-      shooterTab.add("Shooting Mode", 4).withPosition(1, 5).getEntry();
+      shooterTab.add("Shooting Mode", "TEST").withPosition(1, 5).getEntry();
```
azvanderpas: Here the shuffleboard value is a string, but later you send/receive a double. How do you want the value to be stored?

src/main/java/frc/robot/RobotContainer.java
zbrotherton: I'm a little confused on the difference between the two, why they would be called while binding buttons, and if they do work how I think, can the color sensor see our payload?
I'll put them back, I moved them because they were originally in a if that called commands, so I figured the if statement was misplaced, and merged the functionality with the pre-existing if with commands

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -72,7 +76,7 @@
   private NetworkTableEntry ShooterReverted =
       shooterTab.add("Shooter Reverted", false).withPosition(1, 4).getEntry();
```
azvanderpas: Should be "Inverted" rather than "Reverted"

src/main/java/frc/robot/commands/ShooterHeld.java
zbrotherton: True, it's being used as a placeholder for the moment as I don't think the CDS has beam break functionality yet? Correct me if I'm wrong?

zbrotherton: Should we maybe open an issue so we can add the functionality later?

src/main/java/frc/robot/RobotContainer.java
sarango2158: Agreed it is a little weird, just thinking to leave it where Main has it so intake team can be the one to manage where it is called

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -157,224 +172,137 @@ public void updatePID() {
...
-    // SmartDashboard.putNumber("IAccum",KShooterController.getIAccum());
-    // SmartDashboard.putNumber("dist", getDistance());
-    if (DShooterRPM.getDouble(0.0) != shooter_motorController.getEncoder().getVelocity()) {
-      DShooterRPM.setDouble(shooter_motorController.getEncoder().getVelocity());
+    if (DShooterRPM.getDouble(0.0) != flywheelEncoder.getVelocity()) {
+      DShooterRPM.setDouble(flywheelEncoder.getVelocity());
     }
     if (DDistance.getDouble(0.0) != getDistance()) {
       DDistance.setDouble(getDistance());
     }
```
azvanderpas: I don't think the if blocks are needed. We use them for the PID setting because constantly writing new PID values to the motor controllers could cause issues/inefficiencies. But just reading values from encoders/limelight can just be done always.

src/main/java/frc/robot/commands/ShooterHeld.java
sarango2158: Sounds good! Yes my understanding is they don't expose a method for sensor status yet (do those operations internally) so an issue filed to them and a TODO comment here would be good enough

src/main/java/frc/robot/RobotContainer.java
zbrotherton: Works for me

sarango2158: That way you aren't on the hook if the build or simulation fails after moving it, for whatever reason

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -157,224 +172,137 @@ public void updatePID() {
...
-    if (ShooterReverted.getBoolean(false) != shooter_motorController.getEncoder().getInverted()) {
-      shooter_motorController.setInverted(ShooterReverted.getBoolean(false));
+    if (ShooterReverted.getBoolean(false) != flywheelEncoder.getInverted()) {
+      flywheelController.setInverted(ShooterReverted.getBoolean(false));
     }
-    if (DShootingMode.getDouble(0) != aimMode) {
-      aimMode = DShootingMode.getDouble(0);
+    if (DShootingMode.getDouble(0) != aimMode.ordinal()) {
+      setAimMode((int)DShootingMode.getDouble(0));
```
azvanderpas: enums have a .toString() method that could make the Shuffleboard entry easier to parse by the operator. Makes setting the aimMode from shuffleboard a bit trickier though.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -157,224 +172,137 @@ public void updatePID() {
...
-    if (rpm == 0.0) {
-      KShooterController.setReference(0.0, CANSparkMax.ControlType.kVoltage);
-    } else {
-      KShooterController.setReference(rpm, CANSparkMax.ControlType.kVelocity);
+    if(rpm == 0){
+      flywheelPID.setReference(0, CANSparkMax.ControlType.kVoltage);
+      flywheelPID.setIAccum(0);
+    } else{
+    targetRPM = rpm;
+    flywheelPID.setReference(rpm, CANSparkMax.ControlType.kVelocity);
```
azvanderpas: Doesn't look like spotless ran. Make sure you build the code before pushing and creating the PR

src/main/java/frc/robot/commands/ShooterHeld.java
zbrotherton: Issue created #105

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -157,224 +172,137 @@ public void updatePID() {
...
-    switch ((int) aimMode) {
-      case 0: // Case for LOW mode, winds flywheel to preset RPM and adjusts hood to preset
-        // angle
-        adjustHood(Constants.Shooter.LOWAngle);
-        windFlywheel(Constants.Shooter.LOWRPM);
-        currentRPM = Constants.Shooter.LOWRPM;
+    switch (aimMode) {
+      case EJECT: // aimMode used to eject unwanted balls from the shooter
+        adjustHood(aimMode.getAngle());
+        windFlywheel(aimMode.getRPM());
```
azvanderpas:
```suggestion
      case EJECT: // aimMode used to eject unwanted balls from the shooter
      case LOW: // aimMode used to dump into the low goal from ~1ft
      case TARMAC: // aimMode used to shoot into the high goal from ~2ft
      case LAUNCH: // aimMode used to shoot into the high goal from the launchpad
        adjustHood(aimMode.getAngle());
        windFlywheel(aimMode.getRPM());
        break;
```

If multiple cases run the same code, you can just list them like this (notice there's not 'break' between them)

azvanderpas (changes requested):
Consider making constants for the numbers in ShooterSubsystem as well.

src/main/java/frc/robot/commands/ShooterHeld.java
zbrotherton: On the fence about whether it will be used. We might want to run the belt while the shooter is running, instead of having it manual, but there's also cases like #105.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: I forgot to revert that to an int, but it should ideally be referencing our enum

src/main/java/frc/robot/commands/ShooterPressed.java
zbrotherton: Moved to execute, as cargo runs for a set amount of time, it is not just toggled once.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: True, but that brings us the next question: why are we inverting the shooter? Did Jim say?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: Yeah, my primary problem with this was that it makes setting the aimMode hard, ideally we would be setting the enum, but I don't know how to do that with shuffleboard. I think we opened issue #102 for this.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: Will do in the future, for some reason I thought that would happen when we merged to main

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: That's true, I thought switch statements would go on to keep evaluating the others if there was no break, not running the code at the end.

sarango2158 (approved):

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
azvanderpas: Good question. If it isn't needed, then remove it

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -67,11 +67,8 @@
           .add("PID Peak Output Slot ID", Constants.Shooter.kMaxISlotId)
           .withPosition(1, 3)
           .getEntry();
-
-  private NetworkTableEntry ShooterInverted =
-      shooterTab.add("Shooter Inverted", false).withPosition(1, 4).getEntry();
   private NetworkTableEntry DShootingMode =
-      shooterTab.add("Shooting Mode", "TEST").withPosition(1, 5).getEntry();
+      shooterTab.add("Shooting Mode", 4).withPosition(1, 5).getEntry();
```
azvanderpas, line 71: Better to use the AimMode.TEST.ordinal() rather than a number directly

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -155,8 +154,11 @@ private void initCommands() {
       intakeForwardCommand = new IntakeForwardCommand(intakeSubsystem);
       intakeReverseCommand = new IntakeReverseCommand(intakeSubsystem);
     }
-    if (shooterSubsystem != null && limelightSubsystem != null && CDSSubsystem != null) {
-      shooterHeld = new ShooterHeld(shooterSubsystem, limelightSubsystem, CDSSubsystem,true);
+    if (shooterSubsystem != null && CDSSubsystem != null) {
+      shooterHeld =
+          new ShooterHeld(
+              shooterSubsystem, limelightSubsystem, CDSSubsystem, (limelightSubsystem != null));
+      shooterHeld = new ShooterHeld(shooterSubsystem, limelightSubsystem, CDSSubsystem, true);
```
azvanderpas, line 161: Remove this line. Looks like the suggestion feature didn't work right

azvanderpas (approved):
Fix the two minor comments then then it's ready

PR #103, opened 2022-02-19
Two Controller layout 

### LINKED ISSUE #95: Create switchable layouts for button bindings - test and competition
We should be able to quickly and easily change between a controller layout for testing and for competition. Testing layout should expose all our commands/functionality directly to easily troubleshoot. Competition layout should adhere to control scheme we agreed on at the meeting on 2/17

**Competition**
Driver:

- L Bumper - auto-shoot (while held)
- L Trigger - fender shot
- R Bumper - reverse intake
- R Trigger - forward intake
- L joystick - left drive
- R joystick - right drive 

Operator:

- Start button - enable climb
- R Bumper - reverse CDS
- R Trigger - forward CDS
- L joystick - climber control (up and down mapped to axis)

ModBoyex:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -45,7 +46,9 @@
 
   // The robot's subsystems and commands are defined here...
   private static final Joystick driverJoystick = new Joystick(Constants.portNumber0);
+  private static final Joystick driverJoystick2 = new Joystick(Constants.portNumber1);
```
azvanderpas: How about some more specific naming? Something like operatorJoystick

src/main/java/frc/robot/Constants.java
```diff
@@ -86,6 +87,10 @@ public final Boolean isEnabled() {
   // TODO: Replace 0.69 with actual track width in meters and run characterization on real robot
 
   // Encoder constants
+
+  // Controller modes
+  public static final boolean testMode = true;
```
azvanderpas: testMode is pretty vague. Maybe we could set it based on the number of controllers plugged in?

azvanderpas (changes requested):

src/main/java/frc/robot/RobotContainer.java
sarango2158: +1

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -45,7 +46,9 @@
...
   private JoystickButton[] buttons = new JoystickButton[13];
+  private JoystickButton[] buttons2 = new JoystickButton[13];
```
sarango2158: Request better naming here too so it is easy to see which button collections are for which mapping.

sarango2158 (changes requested):
Still a lot more to do for the two different controller layouts we want to have:

- Climber should be on the left joystick on the operator controller for competition mode
- CDS override should be on operator's right bumper and right trigger for competition mode
- Intake should be on driver's right bumper and right trigger for competition mode
- Shoot should be on driver's left bumper and left trigger for competition mode
- Cycling shooter aim modes should only be available for test mode
- Should be able to switch controller layouts without rebuilding and re-deploying. Could be "smart" (detect number of joysticks) or could be a dashboard setting

azvanderpas:
Looks like the Climb base PR included these changes, so this one is now empty. Closing

PR #104, opened 2022-02-19
Climb base

### LINKED ISSUE #86: Add option to enable climb
We want to limit the possibility that our climber is deployed by accident since that will incur a penalty on most of the field. We should have an option to enable climb, and not allow the climber to move up when climb is not enabled. It should be obvious to the user when climb is disabled (indicator on dashboard, physical light, etc.)

### LINKED ISSUE #115: Display "Climb Enabled" on driver dashboard (competition mode)

ModBoyex:


src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -15,35 +16,43 @@
 public class ClimbSubsystem extends SubsystemBase {
   // Put methods for controlling this subsystem
   // here. Call these from Commands.
-
+  private Joystick m_climbJoystick;
   private MotorController m_climbMotorControllerOne;
   private MotorController m_climbMotorControllerTwo;
   private DigitalInput m_limitSwitch;
+  private double climbHight;
```
caseyjbrotherton: typo: climbHeight

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -15,35 +16,43 @@
...
-
-      } else {
-        m_climbMotorControllerOne.getEncoder().setPosition(0);
-        m_climbMotorControllerOne.getPID().setReference(-25, CANSparkMax.ControlType.kPosition);
-      }
+    if (on && !m_limitSwitch.get()) {
+      m_climbMotorControllerOne
+          .getPID()
+          .setReference(
+              climbHight + m_climbJoystick.getRawAxis(Constants.leftJoystickY),
```
caseyjbrotherton: climbHeight is never set to anything besides initialization to zero. Is it means to be keeping track of where we are in the climb?

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -15,35 +16,43 @@
...
   private DigitalInput m_limitSwitch;
+  private double climbHight;
+  private boolean toggleClimb;
 
   // TODO: maybe add servos
 
-  public ClimbSubsystem() {
+  public ClimbSubsystem(Joystick joystick) {
+    m_climbJoystick = joystick;
+    toggleClimb = false;
```
caseyjbrotherton, line 30: toggleClimb doesn't seem to do anything, just set to False here, and toggled in the method.  

Should it be used for some logic to prevent climb, or prevent the command from being scheduled?

Also would be good to have this displayed on shuffleboard at least in a debugging tab.

caseyjbrotherton (changes requested):

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -45,7 +44,9 @@
 
   // The robot's subsystems and commands are defined here...
   private static final Joystick driverJoystick = new Joystick(Constants.portNumber0);
+  private static final Joystick driverJoystick2 = new Joystick(Constants.portNumber1);
```
sarango2158: Carry-over from Controller-Layouts branch, which isn't merged yet/has feedback. Instead of trying to do both at once, could this PR just have the test mode controller layout for climb?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -134,9 +135,8 @@ private void initSubsystems() {
             {
               System.out.println("Climb enabled");
 
-              climbSubsystem = new ClimbSubsystem();
-              climbUPCommand = new ClimbUPCommand(climbSubsystem);
-              climbDOWNCommand = new ClimbDOWNCommand(climbSubsystem);
+              climbSubsystem = new ClimbSubsystem(driverJoystick2);
+              climbCommand = new ClimbCommand(climbSubsystem);
```
sarango2158: Carry-over from Controller-Layouts branch, which isn't merged yet/has feedback. Instead of trying to do both at once, could this PR just have the test mode controller layout for climb? 

For what it's worth here, we're not checking if there are two controllers (e.g. testMode = false) before trying to do this. Could lead to errors or runtime weirdness. Would be good to handle all these cases at once in Controller-Layouts

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -207,8 +207,11 @@ private void configureButtonBindings() {
     }
 
     if (climbSubsystem != null) {
-      buttons[Constants.AButton].whileHeld(climbUPCommand);
-      buttons[Constants.XButton].whileHeld(climbDOWNCommand);
+      if (Constants.testMode) {
```
sarango2158: Carry-over from Controller-Layouts branch, which isn't merged yet/has feedback. Instead of trying to do both at once, could this PR just have the test mode controller layout for climb?

If someone sets testMode = false, we won't have button bindings for climbing. Would be good to do it all at once in Controller-Layouts to account for those cases

sarango2158 (changes requested):

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -15,35 +16,43 @@
...
-    if (on) {
-      if (up) {
-        m_climbMotorControllerOne.getEncoder().setPosition(0);
-        m_climbMotorControllerOne.getPID().setReference(25, CANSparkMax.ControlType.kPosition);
-
-      } else {
-        m_climbMotorControllerOne.getEncoder().setPosition(0);
-        m_climbMotorControllerOne.getPID().setReference(-25, CANSparkMax.ControlType.kPosition);
-      }
+    if (on && !m_limitSwitch.get()) {
```
azvanderpas: Add toggleClimb as a condition here

azvanderpas (changes requested):

src/main/java/frc/robot/RobotContainer.java
ModBoyex: for now, the plan was to just change it to true for false manually depending one if you had one or two contrllers

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -133,10 +134,10 @@ private void initSubsystems() {
           case "ClimbSubsystem":
             {
               System.out.println("Climb enabled");
-
-              climbSubsystem = new ClimbSubsystem();
-              climbUPCommand = new ClimbUPCommand(climbSubsystem);
-              climbDOWNCommand = new ClimbDOWNCommand(climbSubsystem);
+              if (!Constants.oneController) {
+                climbSubsystem = new ClimbSubsystem(driverJoystick2);
+              }
```
azvanderpas: need an else case otherwise the subsystem won't get created if oneController is true

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -207,8 +211,10 @@ private void configureButtonBindings() {
     }
 
     if (climbSubsystem != null) {
-      buttons[Constants.AButton].whileHeld(climbUPCommand);
-      buttons[Constants.XButton].whileHeld(climbDOWNCommand);
+      if (!Constants.oneController) {
+        buttons2[Constants.startButton].whenPressed(
+            new InstantCommand(climbSubsystem::toggleClimbEnable, climbSubsystem));
+      }
```
azvanderpas, line 212: Need an else case

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -45,7 +44,9 @@
 
   // The robot's subsystems and commands are defined here...
   private static final Joystick driverJoystick = new Joystick(Constants.portNumber0);
+  private static final Joystick oporatorJoystick = new Joystick(Constants.portNumber1);
```
caseyjbrotherton: typo

azvanderpas (approved):

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -130,9 +131,10 @@ private void initSubsystems() {
           case "ClimbSubsystem":
             {
               System.out.println("Climb enabled");
-              climbSubsystem = new ClimbSubsystem();
-              climbUPCommand = new ClimbUPCommand(climbSubsystem);
-              climbDOWNCommand = new ClimbDOWNCommand(climbSubsystem);
+              if (!Constants.oneController) {
+                climbSubsystem = new ClimbSubsystem(operatorJoystick);
```
sarango2158, line 134: Because we only have a button mapping for this if there are two controllers, only make this enabled if that second controller is used

sarango2158 (approved):

caseyjbrotherton (approved):
Looks good.  Have two thoughts, can open issues and look at them later:
1.  We should display toggleClimb enabled in shuffleboard. ( This will avoid any confusion during a competition.  Would mark an issue as "competition" )

2.  I am curious how the climbHeight will work.  We aren't waiting for the kPosition PID to finish before incrementing climbHeight again.  Not sure that it will be a problem.  Would suggest sending climbHeight to ShuffleBoard so that you can log it during testing, and examine it.

PR #106, opened 2022-02-19
CDS Auto-Advance

### LINKED ISSUE #87: Auto advance CDS to a position
We should have two locations in the robot that cargo can be:

1. Right behind the cargo stopper wheel
2. One cargo-length beneath that

Each position will be covered by a sensor (beam break or color sensor acting as a beam break). We want the robot to automatically move a ball from the intake, through the "singulator" wheels, and into those positions. If the robot has no cargo, it should move to position 1; if the robot has one cargo, it should move to position 2. This should be conditionally running when the intake is running so the driver and operator do not need to manually operate the CDS, and so we're not running the CDS when it isn't needed.

zainmfjavaid:
Added auto-advance to CDS


src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -13,66 +13,107 @@
...
   
   public int[] getSensorStatus() {
-    int frontStatus = colorSensorOne.getProximity() > 800 ? 1: 0;
-    int backStatus = backBeamBreak.get() ? 1: 0;
-    int[] beamBreakArray = {frontStatus, backStatus};
+    int[] sensorStatuses = colorSensors.getProximities();
+
+    int frontStatus = sensorStatuses[0] > 300 ? 1: 0;
+    int middleStatus = sensorStatuses[1] > 450 ? 1: 0;
+    int backStatus = sensorStatuses[2] > 600 ? 1: 0;
```
azvanderpas: FYI the proximities should correlate with this graph (from [APDS-9151 datasheet](https://www.mouser.com/datasheet/2/678/AVGO_S_A0006247884_1-2574574.pdf), the underlying chip on the color sensors):

image

A lookup type table may be useful at some point, but for now it would be good to make these numbers into constants and/or add comments to describe what they represent.

I'm assuming 300 is used because there is plexiglass in front of that sensor?

```suggestion
    boolean frontStatus = sensorStatuses[0] > 300;
    boolean middleStatus = sensorStatuses[1] > 450;
    boolean backStatus = sensorStatuses[2] > 600;
```

src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) FIRST and other WPILib contributors.
```
azvanderpas: This doesn't seem like it should be a new file. ShooterPressed should already be in main.

src/main/java/frc/robot/commands/CDSReverseCommand.java
```diff
@@ -24,7 +24,8 @@ public CDSReverseCommand(CDSSubsystem CDSSubsystem, ShooterSubsystem shooterSubs
   // Called when the command is initially scheduled.
   @Override
   public void initialize() {
-    mCDSSubsystem.CDSBeltWheelControllerToggle(true);
+    mCDSSubsystem.CDSWheelToggle(true);
+    mCDSSubsystem.CDSBeltToggle(true);
     mShooterSubsystem.runCargo(true, false);
```
azvanderpas: This will cause the robot to shoot, is that whats intended?

src/main/java/frc/robot/commands/OuttakeCommand.java
```diff
@@ -0,0 +1,48 @@
...
+// Open Source Software; you can modify and/or share it under the terms of
+// the WPILib BSD license file in the root directory of this project.
+
+package frc.robot.commands;
+
+import frc.robot.subsystems.CDSSubsystem;
+import frc.robot.subsystems.IntakeSubsystem;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+
+public class OuttakeCommand extends CommandBase {
```
azvanderpas, line 11: Do we need this command and CDSReverseCommand? Seems redundant

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -81,8 +122,34 @@ public void periodic() {
...
     SmartDashboard.putNumber("Ball Count", ballCount);
+
+    // Send ball to setpoint
+    if (!runningCDS) {
+      SmartDashboard.putNumber("Front Sensor Status", sensorStatus[2]);
+      SmartDashboard.putNumber("Middle Sensor Status", sensorStatus[1]);
+      SmartDashboard.putNumber("Back Sensor Status", sensorStatus[0]);
+
+      if (sensorStatus[2] == 1) {  //1 means sensor is activated
+        int nextOpenSensor = getNextOpenSensor();
```
azvanderpas: Keep in mind how many reads you make per cycle to the color sensors. Since the I2C bus is relatively slow it's inefficient to keep accessing it when the data is probably the same

getNextOpenSensor() calls getSensorStatus() again, so right now the proximities are read twice. Consider how you can cut that down to just once. 

Don't have to fix in this PR, but open an issue for it if you don't.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -13,66 +13,107 @@
...
+    // Start at 1 to skip over the centering wheel status 
+    for (int i=0; i < sensorStatus.length-1; i++) {
+      if (sensorStatus[i] == 0) {
+        return i;
+      } 
+    }
+    return -1;
+  }
+
   public void periodic() {
```
azvanderpas, line 152: I don't think this PR should finish before moving this code to a command. Then you can set it up as the default command of the CDS

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -81,8 +122,34 @@ public void periodic() {
 
     // Ball indexing
     int[] sensorStatus = getSensorStatus();
-    int ballCount = sensorStatus[0] + sensorStatus[1];
+    int ballCount = sensorStatus[0] + sensorStatus[1] + sensorStatus[2];
```
azvanderpas: Consider doing this count with a for loop instead so it is easy to add more sensors.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -13,66 +13,107 @@
...
   public int[] getSensorStatus() {
-    int frontStatus = colorSensorOne.getProximity() > 800 ? 1: 0;
-    int backStatus = backBeamBreak.get() ? 1: 0;
-    int[] beamBreakArray = {frontStatus, backStatus};
+    int[] sensorStatuses = colorSensors.getProximities();
+
+    int frontStatus = sensorStatuses[0] > 300 ? 1: 0;
+    int middleStatus = sensorStatuses[1] > 450 ? 1: 0;
+    int backStatus = sensorStatuses[2] > 600 ? 1: 0;
+    int[] beamBreakArray = {backStatus, middleStatus, frontStatus};
```
azvanderpas: The ordering gets a bit confusing since beamBreakArray uses back-to-front order and sensorStatuses uses front-to-back. You can change the order sensorStatuses gets by changing how ColorSensorMuxed is constructed. i.e. 
```
 colorSensors = new ColorSensorMuxed(2,1,0);
```
Will result in getting back-to-front ordering

Or you can change which I2C port the sensors are plugged in to for the same result. (This is possibly better because then the array indices line up with the I2C Port numbers)

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/CDSSubsystem.java
zainmfjavaid: 300 was a number we came up with because of plexiglass interference. We'll add the values we use as constants for this PR and we'll probably end up working on a lookup table or something more scientific going forward.

sarango2158: Would need to cast to int in periodic(), probably in the line where we're adding up the ballCount

src/main/java/frc/robot/subsystems/CDSSubsystem.java
sarango2158: ^ Or less, since we're potentially taking sensors out later on

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -93,9 +160,10 @@ public String getAllianceColor() {
   }*/
 
   public String senseColor() {
-    Color color = colorSensorOne.getColor();
-    double redAmount = color.red;
-    double blueAmount = color.blue;
+
+    Color[] colors = colorSensors.getColors();
+    double redAmount = colors[0].red;
+    double blueAmount = colors[0].blue;
```
sarango2158, line 200: Can we throw a comment here why 0 is the only address we're reading, in case these get moved in the future?

src/main/java/frc/robot/subsystems/Tabs/TabDriveBase.java
```diff
@@ -22,19 +22,19 @@
     // DriveBaseSubsystem object
     private DriveBaseSubsystem mDriveBaseSubsystem;
 
-    private AHRS mGyro;
+    //private AHRS mGyro;
```
sarango2158: Assume these were all removed for testing? If so, uncomment before merging this into main

sarango2158 (changes requested):

src/main/java/frc/robot/subsystems/CDSSubsystem.java
zainmfjavaid: 0 means that we're only sensing color for the sensor under the centering wheels (the sensor leading into the CDS). We could potentially expand this for multiple sensors, but for now, it makes the most sense to only sense color for the first sensor. I'll be sure to add an in-code comment addressing this.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -47,45 +53,140 @@
...
+        ballCount++;
+      }
+    }
+
+    SmartDashboard.putNumber("Ball Count", ballCount);
+
     return beamBreakArray;
   }
 
+  public int getNextOpenSensor(boolean[] sensorSta) {
```
sarango2158: Don't think you need this argument?

src/main/java/frc/robot/subsystems/Tabs/TabDriveBase.java
```diff
@@ -22,19 +22,19 @@
     // DriveBaseSubsystem object
     private DriveBaseSubsystem mDriveBaseSubsystem;
 
-    //private AHRS mGyro;
```
azvanderpas: Revert this so the gyro is removed from this file. It messes with the subsystem disabling and really shouldn't be here.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -90,24 +98,33 @@ public void stopCDSWheel() {
...
-    // Setpoint 2 @ Second beam break
-    int[] sensorStatus = getSensorStatus();
+  public int getNextOpenSensor(boolean[] sensorSta) {
+
+    boolean[] sensorStatus = getSensorStatus();
     
     // Start at 1 to skip over the centering wheel status 
     for (int i=0; i < sensorStatus.length-1; i++) {
-      if (sensorStatus[i] == 0) {
+      if (sensorStatus[i]) {
```
azvanderpas: get rid of sensorStatus and use the sensorSta parameter you made

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -47,45 +53,140 @@
...
+
     return beamBreakArray;
   }
 
+  public int getNextOpenSensor(boolean[] sensorSta) {
+
+    boolean[] sensorStatus = getSensorStatus();
+    
+    // Start at 1 to skip over the centering wheel status 
+    for (int i=0; i < sensorStatus.length-1; i++) {
```
sarango2158, line 144: Doesn't start at 1 like your comment suggests, could return 0

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -47,45 +53,140 @@
...
+    }
+    return -1;
+  }
+
+  public void periodic() {
+    // Color sensing
+    String ballColor = senseColor();
+    SmartDashboard.putString("Ball Color", ballColor);
+    SmartDashboard.putBoolean("Ball Color Match", ballColor == allianceColor);
+    boolean[] sensorStatus = getSensorStatus();
```
sarango2158: Doesn't seem like we need this now that the main logic is in a command. Can at least get rid of the getSensorStatus() because it isn't used, and get rid of pushing the ball color value to the SmartDashboard because senseColor() will already do that

sarango2158 (changes requested):

networktables.ini
```diff
@@ -0,0 +1,10 @@
+[NetworkTables Storage 3.0]
```
azvanderpas: Remove this file from git tracking

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -102,76 +96,76 @@ private void initSubsystems() {
...
-    //Initializes commands based on enabled subsystems
-    if(driveBaseSubsystem != null){
+    // Initializes commands based on enabled subsystems
+    if (driveBaseSubsystem != null) {
       driveBaseTeleopCommand = new DriveBaseTeleopCommand(driveBaseSubsystem);
       driveBaseSubsystem.setDefaultCommand(driveBaseTeleopCommand);
     }
     if(CDSSubsystem != null && shooterSubsystem != null){
-      CDSForwardCommand = new CDSForwardCommand(CDSSubsystem, shooterSubsystem);
+      CDSForwardCommand = new CDSForwardCommand(CDSSubsystem);
```
azvanderpas: Your new command doesn't get instantiated. And shooterSubsystem is no longer a requirement. Try something like this to make the auto advance the default CDS command:

```   
if(CDSSubsystem != null){
      CDSForwardCommand = new CDSForwardCommand(CDSSubsystem);
      CDSSubsystem.setDefaultCommand(new CDSAutoAdvanceCommand(CDSSubSystem));
}
```

azvanderpas (approved):
Fix my last couple comments and this looks good. Nice work!

src/main/java/frc/robot/subsystems/CDSSubsystem.java
zainmfjavaid: Don't know how that got changed. I'll fix that now.

zainmfjavaid: Turns out the comment was outdated but the code was solid. I changed the comment to reflect the current state of things.

src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -71,9 +71,23 @@ public void end(boolean interrupted) {
...
+      //if(i > 0 || m_LimelightSubsystem.calculatePID() == 0.0){
+        m_ShooterSubsystem.runCargo(true,true);
+        m_CDSSubsystem.CDSWheelToggle(false);
+        m_CDSSubsystem.CDSBeltToggle(false);
+        i++;
+        if(i==100){ //Expected to add a 2000ms delay
+          return true;
+       //}
+        }
+    }
```
sarango2158: Merge main one more time -- ShooterPressed doesn't call the CDS directly anymore (will rely on your default command to get the balls into position)

sarango2158 (changes requested):
Looking good, one last change to keep ShooterPressed's changes in main (doesn't use CDS anymore)

sarango2158 (approved):

PR #113, opened 2022-02-22
Simple coordination between the Limelight and the Drivebase

### LINKED ISSUE #85: Don't move side-to-side while shooting the ball
Once the shoot command starts, we want to prevent the robot from getting out of alignment by moving side-to-side. We still want to leave forward/backward as an option to get away from defenders.

zbrotherton:
Closes #85


src/main/java/frc/robot/commands/LimelightAlign.java
```diff
@@ -31,15 +31,14 @@ public void initialize() {}
...
     double adjustment = m_LimelightSubsystem.calculatePID();
-    m_drivebaseSubsystem.setSpeeds(adjustment, adjustment * -1);
+    m_drivebaseSubsystem.arcadeDrive(adjustment);
   }
 
   // Called once the command ends or is interrupted.
   @Override
   public void end(boolean interrupted) {
-    m_drivebaseSubsystem.setSpeeds(0, 0);
+    m_drivebaseSubsystem.arcadeDrive(0);
```
caseyjbrotherton: Is this needed?  if the command is interrupted, should just stay at the heading you are at.

src/main/java/frc/robot/commands/ShooterHeld.java
```diff
@@ -44,6 +49,7 @@ public void initialize() {
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
+    m_DriveBaseSubsystem.arcadeDrive(0);
```
caseyjbrotherton: A quick comment on why you set this to zero would be good.  ( To prevent side to side movement while aligning with the limelight. )

src/main/java/frc/robot/commands/LimelightAlign.java
zbrotherton: To help make sure the limelight doesn't accidentally keep turning the robot, probably not needed but it shouldn't affect anything.

zbrotherton: I'll change it, though

PR #116, opened 2022-02-23
Fixes to the shooter subsystem and lookup table filling

### LINKED ISSUE #111: Add another motor to the Shooter Subsystem
The real robot has two motors on the shooter to spin  the fly wheel. One of these motors needs to be inverted.

zbrotherton:
This branch is intended to take the shooter subsystem itself (not dashboard) most of the way to competition ready.


zbrotherton:
closes #111

caseyjbrotherton:
This is Zane working off of Casey's account btw

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -265,8 +262,12 @@ public void updateSmartDashboard() {
 
   @Override
   public void periodic() {
+    smoothRPM =
+        Constants.Shooter.kA * flywheelEncoder.getVelocity()
```
caseyjbrotherton: is RelativeEncoder.getVelocity() a hardware call?

If so, does it make sense to get it once, and use it for smoothRPM and sending to the dashboard?

src/main/java/frc/robot/Constants.java
```diff
@@ -42,16 +42,16 @@ public final Boolean isEnabled() {
...
     // TODO: Plug real values in for these aimModes
-    LOW(0.0, 0.0) {
+    LOW(3200.0, 0.0) {
       @Override
       public AimModes previous() {
         return values()[values().length - 1];
       }
     },
-    EJECT(0.0, 0.0),
+    EJECT(2500.0, 0.0),
```
caseyjbrotherton, line 51: do we want to have this even lower RPM?  Wondering how quickly we can get sped up and let the opposing ball loose

src/main/java/frc/robot/Constants.java
```diff
@@ -239,20 +239,21 @@ public AimModes previous() {
...
+    public static final double kP = 2.5e-4;
+    public static final double kI = 19e-6;
+    public static final double kD = 0.005;
     public static final double kF = 0.0;
     public static final double kIZone = 0.9;
     public static final double kMaxOutput = 0;
     public static final double kMaxI = 0.9;
     public static final double kMaxISlotId = 0;
     public static final double kMinOutput = 1;
+    public static final double kA = 0.075;
```
caseyjbrotherton, line 257: This feels really small.  The current velocity contributes 0.075 to the SmoothRPM.
The previous velocity contributes ( 0.075)*(1-0.075) = .069 , The one before that contributes ( 0.075)*(1-0.075)^2 = 0.064

We can keep calculating the weights of the previous velocities on the current value...and at 10 values ago, that measurement contributes 0.034.  Which is almost half as impactful as the previous velocity.

To compare, if you set alpha=0.2, then 10 values ago, that velocity contributes 0.02 to the current RPM
If you have output, ,we can look at the impact of different alphas, but I suggest trying at least double what you have.

src/main/java/frc/robot/commands/ShooterEject.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    m_ShooterSubsystem.prime();
+    if (m_ShooterSubsystem.wheelReady()) {
+      i++;
+      m_CDSSubsystem.CDSBeltToggle(false);
+      m_ShooterSubsystem.runCargo(0.65);
```
caseyjbrotherton: This should be a constant.  It is used in multiple commands.  Same with -0.4 below

caseyjbrotherton: Looking at 41-43 with fresh eyes, seems like we shouldn't be setting the speed every loop for any of the commands.  Wonder if that is costly in cycles?

src/main/java/frc/robot/commands/ShooterEject.java
```diff
@@ -0,0 +1,68 @@
...
+    tempMode = m_ShooterSubsystem.getAimMode();
+    i = 0;
+    m_ShooterSubsystem.setAimMode(Constants.AimModes.EJECT);
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    m_ShooterSubsystem.prime();
+    if (m_ShooterSubsystem.wheelReady()) {
```
caseyjbrotherton, line 39: if wheelready or i>0?  

If wheelready falls out of range, then we won't incrmement i during that loop

src/main/java/frc/robot/Constants.java
zbrotherton: All tuning is subject to change, kA will likely turn out to be around 1.5 but that will require testing.

src/main/java/frc/robot/commands/ShooterEject.java
zbrotherton: We do not want to fire if our wheel is not ready, and i should actually be fully reset if the wheel is not ready, come to think of it.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: I don't think this is a hardware call? I've always called this periodically

src/main/java/frc/robot/Constants.java
zbrotherton: All RPMs are subject to change as testing progresses

src/main/java/frc/robot/commands/ShooterEject.java
zbrotherton: I don't know how else I would set this, if not periodically?

I'll turn the speeds into constants

src/main/java/frc/robot/Constants.java
caseyjbrotherton: 0<= kA <= 1.  1.5 won't work

Agree that it is subject to change, suggest that we increase it now before testing, and confirm on your next testing run.

zbrotherton: Sorry, .15 is what I meant to send.

No difference I guess, I'll change it when I can access the code

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
caseyjbrotherton: https://codedocs.revrobotics.com/java/com/revrobotics/relativeencoder#getVelocity()

references JNI ( Java Native Interface ) so I think it is calling a C library to get the velocity.  Unfortunately, RevRobotics code is not open source, so can't tell for sure.

It is not a problem to call it in periodic, but think it best to call it once in one loop if possible.

src/main/java/frc/robot/commands/ShooterEject.java
caseyjbrotherton: once you set the speed of a motor, doesn't it stay until you set it again? 

If this is another hardware call, then it would be good to call it once, if the speed is supposed to stay the same until it is changed.

zbrotherton: The speed is supposed to stay the same until changed, but where do you expect me to change the speed if not in execute()?

caseyjbrotherton: I am suggesting that you change it once in execute.  

https://docs.wpilib.org/en/stable/docs/software/commandbased/commands.html#execution

Execute is called every 20ms.  So, call it the first time to set it, then not after i>0

src/main/java/frc/robot/Constants.java
caseyjbrotherton: .15 sounds good.  10 values after current, the weight is 0.030

caseyjbrotherton (approved):

caseyjbrotherton:
@Majorferguson21 the build fail is likely because of spotless running on the code.  

Should build it locally, and commit again before pushing

PR #120, opened 2022-02-25
Controller layouts

### LINKED ISSUE #95: Create switchable layouts for button bindings - test and competition

ModBoyex:


src/main/java/frc/robot/Robot.java
```diff
@@ -47,6 +52,16 @@ public void robotPeriodic() {
 
     // schedule driverbaseteleop command
     CommandScheduler.getInstance().run();
+    if (Wait == 100) {
+      Wait = 0;
+      AxisCount = DriverStation.getStickAxisCount(Constants.portNumber0);
+      ButtonCount = DriverStation.getStickButtonCount(Constants.portNumber0);
+      AxisCount = DriverStation.getStickAxisCount(Constants.portNumber1);
+      ButtonCount = DriverStation.getStickButtonCount(Constants.portNumber1);
```
sarango2158: You set and then immediately overwrite the Axis and Button counts on port 0 with the values for port 1. Are we doing anything with these values? If so, maybe we should be publishing them separately. If not, we probably don't need to get the axis or button counts for port 0 at all.

One idea we had was to check at portNumber1 that getStickButtonCount > 0 and use that to set the logic for how many controller there are (assumes by default we have at least one).

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -178,44 +181,60 @@ private void initCommands() {
...
-    if (climbSubsystem != null) {
-      if (!Constants.oneController) {
-        buttons2[Constants.startButton].whenPressed(
-            new InstantCommand(climbSubsystem::toggleClimbEnable, climbSubsystem));
+      // Limelight
+      if (limelightAlign != null) {
+        buttons[Constants.startButton].whenPressed(limelightAlign);
+      }
+
+      if (climbSubsystem != null) {}
```
sarango2158, line 265: I think it's fine to leave this empty in case we decide to remap some buttons to allow climb to get tested. But, let's put a comment here for why there is no mapping right now for TestController layout

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -178,44 +181,60 @@ private void initCommands() {
...
+                () -> shooterSubsystem.setAimMode(Constants.AimModes.LOW.ordinal())));
+      }
+
+      if (intakeForwardCommand != null && intakeReverseCommand != null) {
+        buttons[Constants.RTriggerButton].whileHeld(intakeForwardCommand);
+        buttons[Constants.RBumper].whileHeld(intakeReverseCommand);
+      }
+
+      if (climbSubsystem != null) {
+        buttons2[Constants.startButton].whenPressed(climbEnable);
```
sarango2158: Also need an axis for climbCommand

sarango2158 (changes requested):

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -178,44 +181,60 @@ private void initCommands() {
...
+        buttons[Constants.LBumper].whileHeld(shooterHeld);
+        buttons[Constants.LTriggerButton].whileHeld(
+            shooterHeld.beforeStarting(
+                () -> shooterSubsystem.setAimMode(Constants.AimModes.LOW.ordinal())));
+      }
+
+      if (intakeForwardCommand != null && intakeReverseCommand != null) {
+        buttons[Constants.RTriggerButton].whileHeld(intakeForwardCommand);
+        buttons[Constants.RBumper].whileHeld(intakeReverseCommand);
+      }
```
azvanderpas: Take mappings that are the same in both cases out of the if-else block.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -178,44 +181,60 @@ private void initCommands() {
...
+
+      if (climbSubsystem != null) {}
+    } else {
+
+      if (shooterSubsystem != null && shooterHeld != null) {
+        buttons[Constants.LBumper].whileHeld(shooterHeld);
+        buttons[Constants.LTriggerButton].whileHeld(
+            shooterHeld.beforeStarting(
+                () -> shooterSubsystem.setAimMode(Constants.AimModes.LOW.ordinal())));
```
azvanderpas: Lets remove the beforeStarting stuff. Its not the best way to do this. A parameter should get added to shooterHeld for the aimMode @Majorferguson21

src/main/java/frc/robot/Robot.java
ModBoyex: I have changed it to where each one is a different variable and the values are put on Smartdashboard.

src/main/java/frc/robot/Constants.java
```diff
@@ -176,9 +173,13 @@ public AimModes previous() {
   public static final int middleSensorActivation = 450;
   public static final int backSensorActivation = 600;
 
-
   // spotless:off
   // Controller Constants {
+
+    // Controller modes
+    public static final boolean TestController = true;
```
azvanderpas: The naming of the variable is vague. Does true mean we are testing therefore using only 1 controller? oneController seemed a bit more clear. 

At least add comments to describe what true and false mean, but also consider changing the name to make it more clear.

src/main/java/frc/robot/Robot.java
```diff
@@ -18,6 +19,9 @@
   private Command autonomousCommand;
   private RobotContainer robotContainer;
   private TabContainer tabContainer;
+  private int AxisCount;
+  private int ButtonCount;
+  private int Wait;
```
azvanderpas: To keep style consistent make variable names lowerCamelCase

src/main/java/frc/robot/RobotContainer.java
ModBoyex: Comment added

src/main/java/frc/robot/Robot.java
```diff
@@ -47,6 +52,16 @@ public void robotPeriodic() {
 
     // schedule driverbaseteleop command
     CommandScheduler.getInstance().run();
+    if (Wait == 100) {
```
azvanderpas: This code is fine to put here for testing for how to detect the number and type of joysticks, but is shouldn't get into the main branch until you've got a solution that ideally detects joysticks once instead of periodically.

src/main/java/frc/robot/RobotContainer.java
ModBoyex: That is done in the subsystem just like Drivebase

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -178,44 +181,60 @@ private void initCommands() {
...
-    }
+      if (CDSForwardCommand != null && outtakeCommand != null) {
+        buttons[Constants.LTriggerButton].whileHeld(CDSForwardCommand);
+        buttons[Constants.RTriggerButton].whileHeld(outtakeCommand);
+      }
 
-    if (climbSubsystem != null) {
-      if (!Constants.oneController) {
-        buttons2[Constants.startButton].whenPressed(
-            new InstantCommand(climbSubsystem::toggleClimbEnable, climbSubsystem));
```
azvanderpas: This is exactly the same as the ClimbEnable command class you wrote. Why delete it?

src/main/java/frc/robot/commands/ClimbEnable.java
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) FIRST and other WPILib contributors.
```
azvanderpas, line 1: Why not use the InstantCommand you had before?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -131,9 +151,10 @@ private void initSubsystems() {
             }
           case "ClimbSubsystem":
             {
-              if (!Constants.oneController) {
+              if (!Constants.TestController) {
```
azvanderpas: Since you are using the controller detection to set the commands/triggers, you can probably remove this check

azvanderpas (changes requested):

src/main/java/frc/robot/Constants.java
ModBoyex: Now that we have a way to detect how many controllers we have there is no need for those containts.

src/main/java/frc/robot/Robot.java
ModBoyex: I've moved all the stuff in the loop into robotContainer and out of the loop

ModBoyex: soo the wait is gone too

src/main/java/frc/robot/RobotContainer.java
ModBoyex: I got rid of it because I had the cammand

src/main/java/frc/robot/commands/ClimbEnable.java
ModBoyex: I don't remember Why I did that

src/main/java/frc/robot/RobotContainer.java
ModBoyex: I'm going to revert back to this.

azvanderpas:
@AndrewIverson can you verify the button bindings in RobotContainer?

sarango2158 (approved):

sarango2158:
See issue #95 for notes on competition button mapping

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -151,10 +166,10 @@ private void initSubsystems() {
             }
           case "ClimbSubsystem":
             {
-              if (!Constants.TestController) {
+              controllerCheck();
```
azvanderpas: Move this to the top of initSubsystems so other subsystem setups could make use of the variables.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -215,13 +214,15 @@ private void controllerCheck() {
...
-        buttons[Constants.RTriggerButton].whileHeld(intakeReverseCommand);
-      }
 
+    // Intake
+    if (intakeForwardCommand != null && intakeReverseCommand != null) {
+      buttons[Constants.RBumper].whileHeld(intakeForwardCommand);
+      buttons[Constants.RTriggerButton].whileHeld(intakeReverseCommand);
+    }
+
+    if (axisCount1 == 0 && buttonCount1 == 0) {
```
azvanderpas, line 248: You've put button bindings that should be common in both modes here when it should be moved out of the if-else. 

I think the format you want is:
``` 
//common button mappings on first joystick

if ( !second joystick) {
   // first joystick mappings unique to test mode
} else {
   // first joystick mappings unique to competition mode
   // second joystick mappings
}
```

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -215,13 +214,15 @@ private void controllerCheck() {
...
+      
       // Shooter
       if (shooterSubsystem != null && shooterHeld != null) {
         buttons[Constants.backButton].whenPressed(shooterHeld);
```
azvanderpas: #95 says L Bumper

src/main/java/frc/robot/commands/ClimbEnable.java
azvanderpas: Delete unused command class

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -177,46 +213,63 @@ private void initCommands() {
...
         buttons2[Constants.startButton].whenPressed(
-            new InstantCommand(climbSubsystem::toggleClimbEnable, climbSubsystem));
+          new InstantCommand(climbSubsystem::toggleClimbEnable, climbSubsystem));
+      }
+
+      if (CDSForwardCommand != null && outtakeCommand != null) {
+        buttons2[Constants.RBumper].whileHeld(outtakeCommand);
+        buttons2[Constants.RTriggerButton].whileHeld(CDSForwardCommand);
       }
+      System.out.printf("Compitition Mode");
```
azvanderpas: Spelling: Competition 

Also lets print a more specific message string. Something like "Using Competition Two-controller button mappings"
and something similar for Testing mode

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -177,46 +213,63 @@ private void initCommands() {
...
+        CDSSubsystem.setDefaultCommand(new CDSAutoAdvanceCommand(CDSSubsystem));
+      }
 
-    // Limelight
-    if (limelightAlign != null) {
-      buttons[Constants.startButton].whenPressed(limelightAlign);
-    }
+      if (CDSForwardCommand != null && outtakeCommand != null) {
+        buttons[Constants.LTriggerButton].whileHeld(CDSForwardCommand);
+        buttons[Constants.RTriggerButton].whileHeld(outtakeCommand);
```
azvanderpas: RTriggerButton duplicate outtake/intakeReverse above? Work out with CDS/Intake team
#95 says CDS/Intake forward should be RTrigger as well. Work this out with CDS/Intake team

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -177,46 +213,63 @@ private void initCommands() {
...
+      //ClimbSubysystem has no binding because there are not enuf axises
+      if (climbSubsystem != null) {}
+
+      System.out.printf("Testing Mode");
+    } else {
+
+      if (shooterSubsystem != null && shooterHeld != null) {
+        buttons[Constants.LBumper].whileHeld(shooterHeld);
+        buttons[Constants.LTriggerButton].whileHeld(
+                () -> shooterSubsystem.setAimMode(Constants.AimModes.LOW.ordinal()));
```
azvanderpas:
```suggestion
        // Auto Aim Shot
        buttons[Constants.LBumper].whileHeld(
            shooterHeld.beforeStarting(
                () -> shooterSubsystem.setAimMode(Constants.AimModes.AUTO.ordinal()),
                shooterSubsystem));
        // Fender Shot
        buttons[Constants.LTriggerButton].whileHeld(
            shooterHeld.beforeStarting(
                () -> shooterSubsystem.setAimMode(Constants.AimModes.LOW.ordinal()),
                shooterSubsystem));
```

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -92,6 +111,22 @@ public RobotContainer() {
     initializeTrajectories();
   }
 
+  private void controllerCheck() {
```
azvanderpas, line 105: FYI DriverStation has an [isJoystickConnected](https://first.wpi.edu/wpilib/allwpilib/docs/release/java/src-html/edu/wpi/first/wpilibj/DriverStation.html#line.911)​(int stick) method. Seems to check the same things though, so this code is fine as is.

azvanderpas (changes requested):

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -177,46 +213,63 @@ private void initCommands() {
   // it to a {@link
   // edu.wpi.first.wpilibj2.command.button.JoystickButton}.
   private void configureButtonBindings() {
+    controllerCheck();
 
     // Intake
     if (intakeForwardCommand != null && intakeReverseCommand != null) {
       buttons[Constants.RBumper].whileHeld(intakeForwardCommand);
```
AndrewIverson, line 226: Im not sure which direction intakeForward goes. Does it intake the cargo or expel the cargo?
From our controls meeting right bumper was expel cargo, right trigger was intake cargo.

src/main/java/frc/robot/RobotContainer.java
AndrewIverson: Agreed left bumper is while held shoot with limelight aim.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -177,46 +213,63 @@ private void initCommands() {
...
+        CDSForwardCommand = new CDSForwardCommand(CDSSubsystem);
+        CDSSubsystem.setDefaultCommand(new CDSAutoAdvanceCommand(CDSSubsystem));
+      }
 
-    // Limelight
-    if (limelightAlign != null) {
-      buttons[Constants.startButton].whenPressed(limelightAlign);
-    }
+      if (CDSForwardCommand != null && outtakeCommand != null) {
+        buttons[Constants.LTriggerButton].whileHeld(CDSForwardCommand);
```
AndrewIverson: Left trigger should be while held shoot with out lime light auto aim. So the command for the fender shot.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -177,46 +213,63 @@ private void initCommands() {
...
+        buttons[Constants.RTriggerButton].whileHeld(outtakeCommand);
+      }
 
-    if (climbSubsystem != null) {
-      if (!Constants.oneController) {
+      // Limelight
+      if (limelightAlign != null) {
+        buttons[Constants.startButton].whenPressed(limelightAlign);
```
AndrewIverson, line 261: Wrong button, should be left bumper for auto aim while held.

AndrewIverson:
> @AndrewIverson can you verify the button bindings in RobotContainer?

Added comments to incorrect mappings.

src/main/java/frc/robot/commands/ClimbEnable.java
ModBoyex: Command class is used for test code at the moment but we can get rid of that once it is not needed

src/main/java/frc/robot/RobotContainer.java
ModBoyex: Is that not what I'm doing already

src/main/java/frc/robot/RobotContainer.java
ModBoyex: Already fixed

src/main/java/frc/robot/RobotContainer.java
ModBoyex: done

src/main/java/frc/robot/RobotContainer.java
ModBoyex: To fix this I split the button binding for intake, one for competition, and one for testing. The one for testing just uses the bumpers but the one for Comp uses the Buttons binding from issue #95.

src/main/java/frc/robot/RobotContainer.java
ModBoyex: Change Accepted.

src/main/java/frc/robot/RobotContainer.java
ModBoyex: intakeForward takes the ball in, so I inversed it

src/main/java/frc/robot/RobotContainer.java
ModBoyex: Even just for testing mode?

src/main/java/frc/robot/RobotContainer.java
ModBoyex: Is that for Testing mode for Competition mode?

src/main/java/frc/robot/RobotContainer.java
ModBoyex: I thought that the shooter team wanted that for competition but for testing, they wanted it more manual.

src/main/java/frc/robot/RobotContainer.java
ModBoyex: For testing mode, L Bumper is already used by intake but in competition mode, LBumper is the auto-aim shot. Is that not what we wanted?

azvanderpas (approved):

azvanderpas:
Failing git clean check seems to be issue with pipeline and not the code. #

PR #121, opened 2022-02-25
Add build script to copy Autos to the deploy folder on build

azvanderpas:
Pathweaver generates Autonomous paths in the Autos folder. In order to parse them, we need them to get copied into the deploy folder. This addition to the build.gradle script will copy them before building/deploying

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -99,7 +99,7 @@ spotless {
     java {
         target fileTree('.') {
             include '**/*.java'
-            exclude '**/build/**', '**/build-*/**'
+            exclude '**/build/**', '**/build-*/**', '**/deploy/**'
         }
         toggleOffOn()
         googleJavaFormat()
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -11,6 +11,12 @@ targetCompatibility = JavaVersion.VERSION_11
 
 def ROBOT_MAIN_CLASS = "frc.robot.Main"
 
+tasks.register('copyAutos', Copy) {
+    from "Autos"
+    into "src/main/deploy/Autos"
+}
```
sarango2158, line 17: Does the Copy task gracefully handle when the file already exists? The documentation doesn't really seem to say https://docs.gradle.org/current/userguide/working_with_files.html

azvanderpas: It seems to overwrite the files without issues.

sarango2158 (approved):

PR #123, opened 2022-02-25
Git clean check in yaml script

azvanderpas:
Adding a check to make sure building the project doesn't cause the source files to change (i.e. from pushing code that hasn't been built before/had spotless run on it).


caseyjbrotherton (approved):
Looks good.

Another approach for checking is to save the output in a variable:

https://github.com/apache/yetus/blob/2d961971f62f98fe7e15f7979712f292a42f2faa/precommit/src/main/shell/smart-apply-patch.sh#L450

But for the size of our repo, looks good

PR #139, opened 2022-03-05
Add Shuffleboard Commands

Asher123456789:


src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -176,6 +177,7 @@ public void windFlywheel(double rpm) {
     if (rpm == 0) {
       flywheelPID.setReference(0, CANSparkMax.ControlType.kVoltage);
       flywheelPID.setIAccum(0);
+      Shuffleboard.addEventMarker("flyWheel Starts", EventImportance.kHigh);
```
caseyjbrotherton: This line is if the RPM is set to 0.

Would move it before the if.  Maybe it would be good to also capture the RPM?

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -27,6 +30,8 @@ public void toggleIntake(boolean reverse) {
       intakeMotorControllerOne.getSparkMax().set(-Constants.intakeMotorSpeed);
       SmartDashboard.putString("Intake Motor Direction", "Reverse");
       SmartDashboard.putNumber("Intake Motor Speed", -Constants.intakeMotorSpeed);
+      Shuffleboard.addEventMarker("Auto Eject Ran", EventImportance.kHigh);
```
caseyjbrotherton: Only need one event here, not two

src/main/java/frc/robot/commands/ShooterEject.java
```diff
@@ -41,6 +43,7 @@ public void execute() {
         m_CDSSubsystem.CDSBeltToggle(false);
         m_ShooterSubsystem.runCargo(Constants.Shooter.cargoForward);
         m_ShooterSubsystem.setCargoBoolean(true);
+        Shuffleboard.addEventMarker("Fired Cargo", EventImportance.kHigh);
```
caseyjbrotherton: There are other places where cargo is fired.  Would be good to consider if we want to add events to each of the other commands.  

Or, check and see if "setCargoBoolean" is used anywhere else.  Because it might be better to set the "Fired Cargo" event in a method that all three call, as long as it's not used somewhere else

src/main/java/frc/robot/common/hardware/ColorSensorMuxed.java
```diff
@@ -28,6 +30,7 @@ public ColorSensorMuxed(int... ports) {
         i2cPorts.add(p);
         // Initialize each device, only need to keep last object
         sensors = new ColorSensorV3(Port.kOnboard);
+        Shuffleboard.addEventMarker("Sensors Activated", EventImportance.kCritical);
```
caseyjbrotherton: Not sure that this is a good event.  It happens at the beginning of the match for every match.  And only once per match

This is the initialization of the sensors.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
Asher123456789: Combined into one event

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
Asher123456789: Changed accordingly, and RPM has been captured on SmartDashboard

src/main/java/frc/robot/common/hardware/ColorSensorMuxed.java
Asher123456789: Changed it to say sensors initialized, and added a new Shuffleboard event for when they are on and starting to sense color.

src/main/java/frc/robot/commands/ShooterEject.java
Asher123456789: Changed it to where it does that when it fires rather than where it tells the CDS to funnel another ball up.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -173,6 +174,8 @@ public double getVelocityInput() {
   public void windFlywheel(double rpm) {
 
     // Winds Flywheel using PID control to passed rpm
+    Shuffleboard.addEventMarker("FlyWheel Starts", EventImportance.kHigh);
```
caseyjbrotherton: Shuffleboard.addEventMarker("FlyWheel Starts" + rpm, EventImportance.kHigh);

That way, the point in time RPM is logged at the same time as the event.  

There is already a dashboard value for RPM

Asher123456789: Done. I also changed ballCount that was on SmartDashboard to being on Shuffleboard.

PR #144, opened 2022-03-05
Motor controller pid fix

### LINKED ISSUE #97: Autonomous Mode 2: One Ball Fender
- Start at fender
- Shoot preloaded cargo into fender 
- Leave tarmac
- Configurable delay (one value before, one value after)

### LINKED ISSUE #98: Autonomous Mode 3: Two Ball Fender
- Start at tape, line up with an external cargo. Will need two or more positions to avoid conflict with alliance 
- Collect ball while leaving tarmac
- Go back into tarmac 
- Align and shoot two cargo. Configurable between high and low goal (will need to end path at different locations)

tonyn10:


src/main/java/frc/robot/common/hardware/MotorController.java
```diff
@@ -88,33 +82,34 @@ public void setPID() {
     mPIDController.setI(mI);
     mPIDController.setD(mD);
 
-    SmartDashboard.putNumber(mName + " P Value", mP);
-    SmartDashboard.putNumber(mName + " I Value", mI);
-    SmartDashboard.putNumber(mName + " D Value", mD);
-
-    SmartDashboard.putNumber(mName + " FF Value", mFF);
+    updateSmartDashboard(); // post onto smart dashboard
```
azvanderpas: Running updateSmartDashboard() here will overwrite what was just written to the PID variables. I think if you keep the putNumber calls instead it'll fix the issue.

src/main/java/frc/robot/common/hardware/MotorController.java
```diff
@@ -77,7 +71,7 @@ public double getSpeed() {
     return mEncoder.getVelocity();
   }
 
-  public void setPID() {
+  public void activatePID(double[] PID) {
```
azvanderpas: Seems like this method should be private or get merged into setPID

src/main/java/frc/robot/common/hardware/MotorController.java
```diff
@@ -88,33 +82,34 @@ public void setPID() {
...
-      if (SmartDashboard.getNumber(mName + " P Value", mP) != mP) {
-        mP = SmartDashboard.getNumber(mName + " P Value", mP);
+      // TODO: having a sole shuffleboard tab for PID tuning might be beneficical, there's a built
+      // in widget especially for PID
+
+      // check if values were updated in smart dashboard
+      // nothing in it at the start, assign it to value passed in earlier
+      double currentP = SmartDashboard.getNumber(mName + " P value", mP);
+      if (currentP != mP) {
+        mP = currentP;
```
azvanderpas, line 104: This seems functionally the same as it was before. Why do we want to change this?

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -150,44 +153,50 @@ private void initializeRamseteCommands() {
...
                   limelightSubsystem,
                   cdsSubsystem,
                   true), // also has a time delay of 2-3 seconds
               new WaitCommand(Constants.delaytaxi),
-              oneBallRamseteCommand.beforeStarting(
-                  () -> driveBaseSubsystem.resetOdometry(oneBallTrajectory.getInitialPose())));
+              oneBallRamseteCommand
+                  .beforeStarting(
+                      () -> driveBaseSubsystem.resetOdometry(oneBallTrajectory.getInitialPose()))
+                  .andThen(() -> driveBaseSubsystem.stopDriveMotors()));
```
azvanderpas, line 180: I think I noted this before, but if you do the beforeStarting and andThen parts as part of the getRamsetCommand method, then you don't have to add them everywhere (or worry about forgetting them).

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -54,6 +54,9 @@
   private Command threeBallCommand;
   private Command fourBallCommand;
 
+  // amount of time before starting auton
```
azvanderpas, line 57: Units would be good (assuming seconds?)

src/main/java/frc/robot/subsystems/Tabs/TabDriveBase.java
```diff
@@ -39,6 +42,14 @@ public TabDriveBase(DriveBaseSubsystem d) {
...
+    // This is the tab for controlling auton
+    sbWaitTimeSlider =
+        Shuffleboard.getTab("Auton")
+            .add("Max Speed", 1)
+            .withWidget(BuiltInWidgets.kNumberSlider)
+            .withProperties(Map.of("min", 1, "max", 10)) // specify widget properties here
+            .getEntry();
   }
 
   public void periodic() {
```
azvanderpas, line 55: Add @Override annotation

src/main/java/frc/robot/subsystems/Tabs/TabDriveBase.java
```diff
@@ -55,6 +66,9 @@ public void periodic() {
       sbLeftPosition.setDouble(positions[0]); // in meters
       sbRightPosition.setDouble(positions[1]);
 
+      double waitTime = sbWaitTimeSlider.getDouble(1);
+      AutonModes.setWaitTime(waitTime);
```
azvanderpas:
```suggestion
      AutonModes.setWaitTime(sbWaitTimeSlider.getDouble(1));
```

azvanderpas (changes requested):

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -150,44 +153,50 @@ private void initializeRamseteCommands() {
...
+              oneBallRamseteCommand
+                  .beforeStarting(
+                      () -> driveBaseSubsystem.resetOdometry(oneBallTrajectory.getInitialPose()))
+                  .andThen(() -> driveBaseSubsystem.stopDriveMotors()));
 
       twoBallParallel =
           new ParallelDeadlineGroup(
               twoBallRamseteCommand.beforeStarting(
                   () ->
                       driveBaseSubsystem.resetOdometry(
```
sarango2158, line 186: This is the only autonomous mode you have so far that doesn't stop drive motors after resetting odometry. Was that intentional?

src/main/java/frc/robot/subsystems/Tabs/TabDriveBase.java
```diff
@@ -4,7 +4,9 @@
 import edu.wpi.first.networktables.*;
```
sarango2158, line 4: More of a general complaint, but this is the only subsystem that uses TabContainers instead of SmartDashboard and/or Shuffleboard directly. The inconsistency makes it harder to maintain.

sarango2158 (changes requested):

src/main/java/frc/robot/commands/AutonModes.java
tonyn10: The first ramseteCommand only needs to have resetOdometry to reset the heading so that it matches the path.

src/main/java/frc/robot/commands/AutonModes.java
tonyn10: Yes, units are seconds.

src/main/java/frc/robot/common/hardware/MotorController.java
tonyn10: Extra variable to have clarity purposes.

src/main/java/frc/robot/subsystems/Tabs/TabDriveBase.java
tonyn10: If this periodic function is being called in Robot periodic then no need to have it extend SubsystemBase or override.

src/main/java/frc/robot/commands/AutonModes.java
tonyn10: No, I forgot to add that. Thanks for the catch!

src/main/java/frc/robot/subsystems/Tabs/TabDriveBase.java
tonyn10: We might look to disregard these tabs when competition comes around and standardized our output of data with everyone else. For now it's mainly for our testing purposes.

tonyn10:
@azvanderpas Can you see why the git clean check is failing?

azvanderpas:
> @azvanderpas Can you see why the git clean check is failing?

You need to build your code so Spotless will run and commit the changes. Robot.java has whitespace changes.

azvanderpas (dismissed):

sarango2158 (approved):

PR #154, opened 2022-03-10
Configure Shuffleboard Tabs

### LINKED ISSUE #79: Cohesive driver dashboard layout
(related: issue 76)

Asher123456789:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -154,6 +155,7 @@ private void initCommands() {
     if (intakeSubsystem != null) {
       intakeForwardCommand = new IntakeForwardCommand(intakeSubsystem);
       intakeReverseCommand = new IntakeReverseCommand(intakeSubsystem);
+      Shuffleboard.addEventMarker("Reverse Intake Activated", EventImportance.kLow);
```
sarango2158: This isn't a reverse intake event, just creating the instance of the IntakeReverseCommand class we'll use. Should probably have this called inside the initialize() method of that command instead

src/main/java/frc/robot/commands/ShooterEject.java
```diff
@@ -3,6 +3,8 @@
 
 package frc.robot.commands;
 
+import edu.wpi.first.wpilibj.shuffleboard.EventImportance;
```
sarango2158: Looks like this isn't used?

src/main/java/frc/robot/common/hardware/ColorSensorMuxed.java
```diff
@@ -57,6 +59,7 @@ private boolean setI2cPort(int port) {
     for (int p : i2cPorts) {
       if (setI2cPort(p)) {
         colors[i] = sensors.getColor();
+        Shuffleboard.addEventMarker("Sensors Activated", EventImportance.kCritical);
```
sarango2158: Should probably add this same event to getProximities() as well? Since both getColors() and getProximities() will read from the sensors. 

Separate thought -- does addEventMarker record a timestamp? It would be really great to have a sense of how often we were polling the sensors. Does not have to be in this PR.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -71,6 +72,8 @@ public void CDSWheelToggle(boolean reverse) {
     if (reverse) {
       CDSWheelControllerOne.getSparkMax().set(-Constants.CDSWheelControllerSpeed);
       CDSWheelControllerDirection.setString("Reverse");
+      ShuffleboardTab OperatorView = Shuffleboard.getTab("CDS Speed");
```
sarango2158, line 75: Should the other events go into a specific tab as well? Perhaps they could all be on OperatorView

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -23,10 +28,13 @@ public IntakeSubsystem() {
   }
 
   public void toggleIntake(boolean reverse) {
+    ShuffleboardTab OperatorView = Shuffleboard.getTab("Intake Speed");
+    OperatorView.add("Intake Wheel Speed", Constants.intakeMotorSpeed).withWidget(BuiltInWidgets.kDial);
```
sarango2158, line 32: Should we care whether intake is running forward or reverse for this display (positive or negative value)? Since it is a constant value and we're not providing a way to overwrite it, I'm not sure this is helpful on its own

sarango2158 (changes requested):

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
caseyjbrotherton: @Majorferguson21 had asked for this to be a value instead of a boolean or string.  Thoughts, Zane?

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -133,8 +136,8 @@ public void stopCDSWheel() {
         ballCount++;
       }
     }
-
     SmartDashboard.putNumber("Ball Count", ballCount);
+    Shuffleboard.addEventMarker("" + ballCount, EventImportance.kNormal);
```
caseyjbrotherton: Let's remove this, as we already log the ball count.  Should the ball count be moved to operator or driver tab? That would be a good change here.

src/main/java/frc/robot/common/hardware/ColorSensorMuxed.java
caseyjbrotherton: When shuffleboard is recording, it does log a timestamp.
Not sure that we should write an event for every time the sensor is active though. 
As we are sending the color to shuffleboard.

caseyjbrotherton (changes requested):

src/main/java/frc/robot/subsystems/CDSSubsystem.java
caseyjbrotherton: All of the "OperatorView" tabs should be the same tab.  
i.e:

ShuffleboardTab OperatorView = Shuffleboard.getTab("Operator View")

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
zbrotherton: We should probably make sure this is a NetworkTableEntry, so it can be updated as the direction of the intake changes.
A value is best, so we can show the direction the intake is spinning on a slider-type widget using only 1 entry and only 1 widget.

caseyjbrotherton: Looking again, I see what you mean Siana. We should putNumber after line 36 with -Constants.intakeMotorSpeed, and after line 41 with Constants.intakeMotorSpeed.  That will show the changes when the motor is reversed.

Also, let's put the "add" in the constructor of the subsystem and put a putNumber call here.

src/main/java/frc/robot/RobotContainer.java
Asher123456789: Deleted and moved to the proper location.

src/main/java/frc/robot/commands/ShooterEject.java
Asher123456789: True, I think I had something in there originally and needed those but now I have deleted them as they are no longer necessary.

src/main/java/frc/robot/common/hardware/ColorSensorMuxed.java
Asher123456789: As it has been a bit since this PR was made, many features have been added or removed since this. There is no longer a shuffleboard event anywhere in ColorSensorMuxed, and I have added a similar feature to CDSSubsystem when the sensor uses getProximities.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
Asher123456789: This has already been addressed in the latest version of main, as this PR is quite old.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
Asher123456789: It has been deleted in the latest version of main and moved to only activate when debug mode is enabled.

azvanderpas:
PR is fairly stale. Closing for now. Feel free to reopen/resubmit after updating

PR #155, opened 2022-03-10
Climb base

ModBoyex:


loipiop224 (dismissed):
You're good to go

src/main/java/frc/robot/Constants.java
```diff
@@ -20,7 +20,7 @@
     // Change booleans to disable a subsystem in RobotContainer
     // spotless:off
     DriveBaseSubsystem(true),
-    CDSSubsystem      (true),
+    CDSSubsystem      (false),
```
azvanderpas: Undo

src/main/java/frc/robot/commands/ClimbEnable.java
```diff
@@ -0,0 +1,48 @@
...
+  public void execute() {}
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {}
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    return false;
```
azvanderpas: Since you want this to run once and then be done, I think you want to set this return value to true. Otherwise the whenPressed trigger will cause the execute method to keep running forever.

azvanderpas (changes requested):
Test on robot before allowing into main

src/main/java/frc/robot/Constants.java
ModBoyex: we were trying to test and CDS was giving us an error when testing, so we should probably look into that, but I did change it back.

src/main/java/frc/robot/commands/ClimbEnable.java
ModBoyex: fixed

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -67,55 +71,103 @@ public ClimbSubsystem(Joystick joystick) {
...
 
-    sbclimberpositionTwo =
-        climberTab.add("Climber position 2", 0).withSize(2, 2).withPosition(2, 4).getEntry();
+    m_climbMotorControllerOne.getPIDCtrl().setP(sbClimbOneP.getDouble(0));
+    m_climbMotorControllerOne.getPIDCtrl().setI(sbClimbOneI.getDouble(0));
+    m_climbMotorControllerOne.getPIDCtrl().setD(sbClimbOneD.getDouble(0));
+
+    m_climbMotorControllerTwo.getPIDCtrl().setP(sbClimbTwoP.getDouble(0));
+    m_climbMotorControllerTwo.getPIDCtrl().setI(sbClimbTwoI.getDouble(0));
+    m_climbMotorControllerTwo.getPIDCtrl().setD(sbClimbTwoD.getDouble(0));
```
azvanderpas: Unnecessary since MotorController sets the defaults now, and updating is already setup in periodic

src/main/java/frc/robot/commands/ClimbCommand.java
```diff
@@ -24,7 +25,9 @@ public void initialize() {}
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
+
     if (m_subsystem.getclimbingmode()) {
+      m_subsystem.resetTargetedHeight();
```
azvanderpas, line 30: This seem like it should go in ClimbEnable instead.

azvanderpas (dismissed):

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
ModBoyex: Fixed that (I blame Jim for that one)

src/main/java/frc/robot/commands/ClimbCommand.java
sarango2158: I agree with Adam, makes more sense to reset when climb is enabled instead of conditionally here. What's the thought for keeping it here?

ModBoyex: fixed (forgot to git rid of the old one)

sarango2158 (approved):

PR #159, opened 2022-03-10
Reverting ff changes to prepare for open house

zbrotherton:
Calculating an FF based off of characterization is going to take time, and I accidentally left some of those changes in with latest hotfix.

### DIFF
--- a/src/main/java/frc/robot/subsystems/ShooterSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/ShooterSubsystem.java
@@ -7,6 +7,7 @@ package frc.robot.subsystems;
 import com.revrobotics.CANSparkMax;
 import com.revrobotics.RelativeEncoder;
 import com.revrobotics.SparkMaxPIDController;
+import edu.wpi.first.math.controller.SimpleMotorFeedforward;
 import edu.wpi.first.networktables.NetworkTableEntry;
 import edu.wpi.first.networktables.NetworkTableInstance;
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -64,6 +67,9 @@ public ShooterSubsystem() {
         new MotorController("Flywheel", Constants.Shooter.shooterID, Constants.Shooter.kPIDFArray);
     flywheel2Controller = new MotorController("Flywheel 2", Constants.Shooter.shooter2ID);
     flywheelPID = flywheelController.getPIDCtrl();
+    flywheelFF =
+        new SimpleMotorFeedforward(
+            Constants.Shooter.kSg, Constants.Shooter.kVg, Constants.Shooter.kAg);
```
caseyjbrotherton, line 72: This is confusing as flywheelFF isn't actually used.

caseyjbrotherton (dismissed):
pushing in, as Zane is in the robot lab during school trying to fix main

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: It's something I was experimenting with when I realized I broke main on accident
Will not have an effect

caseyjbrotherton (approved):

PR #161, opened 2022-03-11
Shooter auton and one ball mode

tonyn10:
One ball mode with shooter working in auton. Still testing two ball mode, just love to get a review and some feedback on what I have so far (I made a new branch called AutonTwoBall to continue testing.)


src/main/java/frc/robot/Robot.java
```diff
@@ -36,10 +36,10 @@ public void robotInit() {
...
     // TODO: change to correct paths
     chooser.setDefaultOption("Taxi", "taxi");
-    chooser.addOption("One Ball", "oneball");
-    chooser.addOption("Two Ball", "twoball");
-    chooser.addOption("Three Ball", "threeball");
-    chooser.addOption("Four Ball", "fourball");
+    chooser.addOption("One Ball", "one ball");
+    chooser.addOption("Two Ball", "two ball");
+    chooser.addOption("Three Ball", "three ball");
+    chooser.addOption("Four Ball", "four ball");
```
azvanderpas, line 42: An enum might be help prevent mismatches in the future.

src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -55,7 +55,8 @@ public void execute() {
         }
         i++;
       }
-    } else {
+    } else if (i == 0) {
+      // when wheel is not ready and i is still 0
```
azvanderpas, line 59: @Majorferguson21 this logic fixing the autonomous shooting implies that wheelReady() is jittery (doesn't stay true for long enough). PID turning will possibly make this better, but since we are shooting with good accuracy regardless I think this will work for competition. Might come into play more if/when we do longer range shots.

azvanderpas (approved):

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
```diff
@@ -297,8 +302,30 @@ public void setSpeeds(double left, double right) {
...
+      m_motorControllers[Constants.driveRightRear].setInverted(true);
+      m_motorControllers[Constants.driveLeftFront].setInverted(false);
+      m_motorControllers[Constants.driveLeftRear].setInverted(false);
+    } else {
+      m_motorControllers[Constants.driveRightFront].setInverted(false);
+      m_motorControllers[Constants.driveRightRear].setInverted(false);
+      m_motorControllers[Constants.driveLeftFront].setInverted(true);
+      m_motorControllers[Constants.driveLeftRear].setInverted(true);
+    }
+  }
```
azvanderpas, line 319: I think possibly the missing piece to running a vector backwards is negating the getPose value from the gyro. Though the reverse vector option in pathweaver is probably going to work better either way.

PR #163, opened 2022-03-12
Auton remodel

tonyn10:
AutonModes was a bit sloppy. Added enum for organization and condensed code by using varargs. Also added the ability to use test mode (just driving in autonomous) which will help with figuring out pathweaver paths.


src/main/java/frc/robot/Robot.java
```diff
@@ -34,13 +34,19 @@ public void robotInit() {
...
+    chooser.addOption("One Ball", Constants.Auton.ONEBALL);
+    chooser.addOption("Two Ball", Constants.Auton.TWOBALL);
+    chooser.addOption("Three Ball", Constants.Auton.THREEBALL);
+    chooser.addOption("Four Ball", Constants.Auton.FOURBALL);
+    chooser.addOption("Five Ball", Constants.Auton.FIVEBALL);
+    chooser.addOption("Test Mode", Constants.Auton.TEST);
+
+    SmartDashboard.putData(
+        "Auto Mode",
+        chooser); //  TODO: find a way to put it into desired specific named tabs such as "Auton"
```
caseyjbrotherton, line 49: https://github.com/caseyjbrotherton/ShuffleBoardExamples/blob/main/src/main/java/frc/robot/subsystems/ExampleSubsystem.java#L38

Has examples about tabs.  Would suggest talking with @Asher123456789 and @Majorferguson21 as they were looking at moving stuff to an "Operator" tab and a "Driver" tab for competition.

( Would recommend doing that after this is merged.

src/main/java/frc/robot/Constants.java
```diff
@@ -137,12 +137,31 @@ public AimModes previous() {
...
+  public enum Auton {
+    TAXI("paths/TaxiOut.wpilib.json"),
+    ONEBALL("paths/TaxiOutFromFender.wpilib.json"),
+    TWOBALL("paths/TaxiOutToGrabBall.wpilib.json", "paths/GoBackIntoFender.wpilib.json"),
+    THREEBALL,
+    FOURBALL,
+    FIVEBALL,
+    TEST(
+        "paths/FlippedVectorTest1.wpilib.json",
+        "paths/FlippedVectorTest2.wpilib.json"); // change according to what path you want to test
```
azvanderpas: This is definitely a spot where spotless makes things less readable. You can disable it for chucks of code like this:
```
// spotless: off
<code you want to manual format>
// spotless: on 
```

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -27,47 +27,49 @@
...
   private Command twoBallCommand;
   private Command twoBallParallel;
+
   private Command threeBallCommand;
   private Command fourBallCommand;
+  private Command fiveBallCommand;
+
+  private Command testCommand; // for testing miscellaneous: for example single ramsete commands
 
   // amount of time before starting auton
```
sarango2158: Add units to comment

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -166,29 +170,20 @@ private void initializeCommandGroups() {
...
 
       twoBallParallel =
           new ParallelDeadlineGroup(
-              twoBallRamseteCommand[0].beforeStarting(
-                  () ->
-                      driveBaseSubsystem.resetOdometry(
-                          twoBallTrajectory[0].getInitialPose())), // go out to get ball
+              twoBallRamseteCommands[0], // travel to get ball
               new IntakeForwardCommand(intakeSubsystem, cdsSubsystem)
                   .andThen(() -> driveBaseSubsystem.stopDriveMotors()));
```
azvanderpas: It seems like this could be added to the ramsete command and have the same effect. Same with below in the twoBallCommand. stopDriveMotors() can probably be added to every ramsete command in your getRamsetCommand method. I don't see any cases where you wouldn't want to stop after a trajectory.

src/main/java/frc/robot/Constants.java
```diff
@@ -137,12 +137,31 @@ public AimModes previous() {
...
+
+    private String paths[];
+
+    private Auton() {
+      paths = null;
+    }
+
+    private Auton(String... paths) {
+      this.paths = paths;
+    }
```
azvanderpas: Just an FYI, since paths is already optional, you don't need to make the Auton() method with no parameters. Fine as is though.

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -76,25 +78,21 @@ public AutonModes(DriveBaseSubsystem d, IntakeSubsystem i, CDSSubsystem c) {
...
     this.shooterSubsystem = s;
     this.limelightSubsystem = l;
     this.cdsSubsystem = c;
     this.intakeSubsystem = i;
 
-    allSubsystemsEnabled = true;
-
-    // parameters are route names (names based on pathweaver, they are located under Auton file)
-    initializeTrajectories();
+    shooterEnabled = true;
```
sarango2158, line 98: If this is false, we'll have a handful of auton modes with no ramsete commands that are still choosable. We probably don't need this "shooterEnabled" logic for competition at all, but at minimum we should control this in a more visible way than a private variable in this class. Publish to dashboard, tie to modular subsystem, etc.

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -103,55 +101,61 @@ public AutonModes(
...
 
   private void initializeRamseteCommands() {
-    taxiRamseteCommand = getRamseteCommand(taxiTrajectory);
+    taxiRamseteCommand =
+        getRamseteCommands(getTrajectories(Constants.Auton.TAXI.getPaths()))[
+            0]; // only has one trajectory so only one ramsete command (first element of array)
 
-    if (allSubsystemsEnabled) {
-      oneBallRamseteCommand = getRamseteCommand(oneBallTrajectory);
+    if (shooterEnabled) {
```
sarango2158, line 155: Related to previous comment on 95. We probably don't need this for competition, so ideally we'd get rid of this. At minimum, it should be easier for the user to remember to set and disable (tied to modular subsystem or dashboard).

sarango2158 (changes requested):

sarango2158 (dismissed):

sarango2158 (approved):

PR #171, opened 2022-03-12
Shuffleboard

### LINKED ISSUE #76: Cohesive debug dashboard layout

### LINKED ISSUE #79: Cohesive driver dashboard layout

### LINKED ISSUE #176: CC2 has CommandScheduler Loop Overruns when two joysticks are plugged in

zbrotherton:
Needs testing before pull


zbrotherton:
Expected to fix #176 (Climber periodic makes too many calls to hardware and SmartDashboard calls in main)

src/main/java/frc/robot/Robot.java
```diff
@@ -36,7 +36,7 @@ public void robotInit() {
     // TODO: Put commands here
 
     usbCamera = CameraServer.startAutomaticCapture();
-    usbCamera.setResolution(320, 240);
+    usbCamera.setResolution(240, 320);
```
azvanderpas: This seem to break simulation in Windows, though it apparently doesn't complain on the pipeline Linux machine. 
Wrap with if (isReal()) {}

src/main/java/frc/robot/commands/ClimbEnable.java
```diff
@@ -27,7 +27,7 @@ public void initialize() {
     m_subsystem.climbEnable();
     m_subsystem.resetTargetedHeight();
 
-    if (m_subsystem.getclimbingenable() == true) {
+    if (m_subsystem.getclimbingmode() == true) {
```
azvanderpas: 'Enabled' is actually the better naming scheme for documentation. 'Mode' is ambiguous about what the true/false values mean. Consider something like 'isClimbingEnabled' instead.

'== true' is unnecessary as well

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -63,20 +79,30 @@ public CDSSubsystem() {
   public void CDSToggleAll(boolean reverse) {
     if (reverse) {
       CDSWheelControllerOne.set(-Constants.CDSWheelControllerSpeed);
-      SmartDashboard.putString("CDS Wheel Direction", "Reverse");
-      SmartDashboard.putNumber("CDS Wheel Speed", -Constants.CDSWheelControllerSpeed);
+      DCDSSpeed.setDouble(-1);
```
sarango2158, line 80: The name of this entry implies we want the actual speed, but we are giving it a static value that could differ from wheel speed. Maybe use Constants.CDSWheelControllerSpeed instead

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -63,20 +79,30 @@ public CDSSubsystem() {
...
 
       CDSBeltController.set(-Constants.CDSBeltSpeed);
-      SmartDashboard.putString("CDS Belt Direction", "Reverse");
-      SmartDashboard.putNumber("CDS Belt Speed", -Constants.CDSBeltSpeed);
+      if (Constants.DebugMode) {
+        SmartDashboard.putString("CDS Belt Direction", "Reverse");
+        SmartDashboard.putNumber("CDS Belt Speed", -Constants.CDSBeltSpeed);
+      }
     } else {
+      DCDSSpeed.setDouble(1);
```
sarango2158, line 92: Same feedback as on line 82. Tie to actual constant speed value

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -86,21 +112,29 @@ public void CDSWheelToggle(boolean reverse) {
...
-      SmartDashboard.putNumber("CDS Wheel Speed", Constants.CDSWheelControllerSpeed);
+      if (Constants.DebugMode) {
+        SmartDashboard.putString("CDS Wheel Direction", "Forward");
+        SmartDashboard.putNumber("CDS Wheel Speed", Constants.CDSWheelControllerSpeed);
+      }
     }
   }
 
   public void CDSBeltToggle(boolean reverse) {
+    DCDSSpeed.setDouble(-1);
```
sarango2158, line 121: See feedback on 82

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -86,21 +112,29 @@ public void CDSWheelToggle(boolean reverse) {
...
       CDSBeltController.set(-Constants.CDSBeltSpeed);
-      SmartDashboard.putString("CDS Belt Direction", "Reverse");
-      SmartDashboard.putNumber("CDS Belt Speed", -Constants.CDSBeltSpeed);
+      if (Constants.DebugMode) {
+        SmartDashboard.putString("CDS Belt Direction", "Reverse");
+        SmartDashboard.putNumber("CDS Belt Speed", -Constants.CDSBeltSpeed);
+      }
 
     } else {
+      DCDSSpeed.setDouble(1);
```
sarango2158, line 130: See feedback on 82

src/main/java/frc/robot/subsystems/CDSSubsystem.java
zbrotherton: Naming is a bit ambiguous, but the intention is to just show direction in a manner that is clear on the dashboard, the actual speed isn't necessary

src/main/java/frc/robot/commands/ClimbEnable.java
zbrotherton: climbing enable isn't something I deleted, I was confused as to what climbing mode meant, but if enabled is what was meant there I'll change it

caseyjbrotherton (dismissed):
Adding to main for last precompetition system test

caseyjbrotherton (approved):

PR #182, opened 2022-03-14
Prepare for monday driver testing

### LINKED ISSUE #173: Tarmac shot in CC2 is at too high of rpm
Is currently at 2600 rpm should be 2550 rpm.

### LINKED ISSUE #174: CC2 Drive Base does not set correct current limit
The motor controllers currently default to 40amp current limit, the drive base should be set to 50 amp current limit on all  motors

### LINKED ISSUE #175: Motor Controller should not set openloop ramp rate by default.
Open loop ramp rate is only really useful in certain situations like on the drive base. We don't want all motors getting a .2 second delay for full power.

### LINKED ISSUE #177: CC2 arcade drive turn rate is wrong
arcade drive for operator control turning should be multiplied by .85.

### LINKED ISSUE #178: CC2 Stopper wheel doesn't run while intaking
This can occasionally cause a cargo to be ejected from the shooter while we are intaking the second cargo

### LINKED ISSUE #179: CC2 Climber Network Table Entry Usage
There is no reason to use a network table entry in drive subsystem for teleop speed, just use a private double in drivebase for teleopSpeed.

### LINKED ISSUE #180: CC2 Climber needs to always be intilized.
Currently the climber subsystem is only initialized if there are two joysticks plugged into the driver system. Depending on setup order we might have the robot on and then check if the joysticks are plugged in. The climb subsystem should always be enabled.

### LINKED ISSUE #181: Button mapping is dependent on controller count
In ConfigureButtonBindings() 
We map different commands dependent on  how many joysticks are plugged into the computer when the roborio first boots. That's not really a great way to do button mapping. For competition we need a static button mapping that isn't dependent on if the controllers are plugged in.

Consider the situation: 
One controller gets unplugged we place the robot on the field and turn it on, some one notices the cable is unplugged and plugs it back in. 
Problem:
Things are now bound differently and the user/driveteam doesn't know things are bound differently until competition starts and things are not on the controls they expect.

zbrotherton:


AndrewIverson (dismissed):
Looks good.

AndrewIverson:
Will merge after testing today if it goes well.

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
```diff
@@ -217,24 +232,23 @@ public void periodic() {
...
   }
 
   // Normal Arcade Drive
   public void arcadeDrive() {
     // Note: -0.85 to accomodate comfort of driver (sensitivity)
     m_differentialDrive.arcadeDrive(
-        m_driverJoystick.getRawAxis(Constants.leftJoystickY) * -(teleopSpeed.getDouble(100) / 100),
-        m_driverJoystick.getRawAxis(Constants.rightJoystickX),
+        m_driverJoystick.getRawAxis(Constants.leftJoystickY) * -driveBaseSpeed,
+        m_driverJoystick.getRawAxis(Constants.rightJoystickX) * -Constants.driveBaseTurnRate,
```
azvanderpas: I don't think the multiplier should be negative anymore.

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
```diff
@@ -217,24 +232,23 @@ public void periodic() {
...
         true);
     // joystick has y-axis flipped so up is negative why down is positive
   }
 
   // Arcade Drive where you can only move forwards and backwards for testing
   public void arcadeDrive(double rotation) {
-    m_differentialDrive.arcadeDrive(
-        -0.85 * m_driverJoystick.getRawAxis(Constants.leftJoystickY), rotation);
+    m_differentialDrive.arcadeDrive(m_driverJoystick.getRawAxis(Constants.leftJoystickY), rotation);
```
azvanderpas: Why not add the multiplier here?

AndrewIverson (dismissed):

AndrewIverson (approved):

PR #186, opened 2022-03-15
Auton two ball

tonyn10:
Changed the path to start at tape line and make it drive more centered to fender, still needs testing. Added shuffleboard for auton data in Robot class.


src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -19,7 +20,7 @@
 import java.io.IOException;
 import java.nio.file.Path;
 
-public class AutonModes {
+public class AutonModes extends SubsystemBase {
```
azvanderpas: Why do we want it to be a subsystem?

azvanderpas (dismissed):

src/main/java/frc/robot/commands/AutonModes.java
tonyn10: I'll fix that, it was a discontinuous idea.

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -192,13 +192,12 @@ private void initializeCommandGroups() {
 
       twoBallCommand =
           new SequentialCommandGroup(
-              new DeployIntake(intakeSubsystem, cdsSubsystem),
+              // new DeployIntake(intakeSubsystem, cdsSubsystem),
```
sarango2158, line 195: I think we still want to deploy the intake in autonomous, was this just for testing?

sarango2158 (changes requested):

src/main/java/frc/robot/commands/AutonModes.java
tonyn10: The parallel command that has combined intake and cds will deploy it.

sarango2158 (approved):

PR #187, opened 2022-03-15
Only update I2C sensors if data is stale (100 ms default period)

azvanderpas:


src/main/java/frc/robot/common/hardware/ColorSensorMuxed.java
```diff
@@ -52,31 +62,35 @@ private boolean setI2cPort(int port) {
...
-        colors[i] = sensors.getColor();
-      } else {
-        DriverStation.reportError("Failed to get color from sensor on I2C port " + p, false);
-        colors[i] = new Color(0, 0, 0);
+    if (Timer.getFPGATimestamp() - lastColorRead > sensorPeriodInSeconds) {
+      int i = 0;
+      for (int p : i2cPorts) {
+        if (setI2cPort(p)) {
+          colors[i] = sensors.getColor();
+          lastColorRead = Timer.getFPGATimestamp();
```
caseyjbrotherton: Assume we set this here instead of at line 66 because we want to make sure that setI2cPort returns successful for one of the I2c ports.  We are calling getFPGATimestamp possibly three times though.

Not sure if we should be concerned about that.  Thoughts?  

Looks like getFPGATimestamp is fairly quick according to McCaskle: 

 https://www.chiefdelphi.com/t/velocity-pid-programming-terms-dependent-on-time/126066/32?u=casey

sarango2158 (approved):

PR #188, opened 2022-03-16
CDS fixes

Ariya3:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -195,14 +195,14 @@ private void configureButtonBindings() {
     // Intake / CDS
     if (outtakeCommand != null) {
       // spits ball out
-      buttons[Constants.RBumper].whileHeld(outtakeCommand);
+      buttons[Constants.AButton].whileHeld(outtakeCommand);
```
azvanderpas: Keep this at RBumper

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -40,79 +40,95 @@ public void initialize() {}
...
+        SmartDashboard.putBoolean("Back Sensor Status", sensorStatus[0]);
+
+        // Calls Auto Eject
+        if (!ejectRunning) {
+          // Checks if conditions for ejection are met:
+          // A ball count of over 2 OR ball color is wrong and test mode is off (meaning ball color
+          // shouldn't be disregarded)
+          // TODO This is somehow ejecting any ball regardless of color
+
+          // Uncomment Once Done
```
azvanderpas: Fixed, so we can delete this comment

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -40,79 +40,95 @@ public void initialize() {}
...
-            int nextOpenSensor = CDSSubsystem.getNextOpenSensor(sensorStatus);
-            SmartDashboard.putNumber("Open Sensor Index", nextOpenSensor);
-            if (nextOpenSensor != -1) {
-              // There is an open sensor avaliable, run CDS
-              runningCDS = true;
-              sensorIndex = nextOpenSensor;
-              CDSSubsystem.CDSBeltToggle(false);
-              CDSSubsystem.setReady(false);
-            }
+      if (msCurrent >= runInterval) {
```
azvanderpas: With the updates in #187 we should be able to remove this check

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -40,79 +40,95 @@ public void initialize() {}
...
+            intakeSubsystem.toggleIntake(true);
+            CDSSubsystem.CDSWheelToggle(true);
+            ejectRunning = true;
+            SmartDashboard.putBoolean("Eject Running", ejectRunning);
           }
         } else {
-          // Check if ball has reached sensor, stop if it has
-          SmartDashboard.putNumber("Open Sensor Index", sensorIndex);
-          if (sensorStatus[sensorIndex]) {
+          if (msCurrent >= msDelay) {
```
azvanderpas: !sensorStatus[2] may work better as the termination condition. Maybe with the timeout as well

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -40,79 +40,95 @@ public void initialize() {}
...
         }
+
+        // Only run auto advance if auto ject is not running
+        if (!ejectRunning) {
+          SmartDashboard.putBoolean("CDS Running", runningCDS);
+          if (!runningCDS) {
+            // Send ball to sensor
+            SmartDashboard.putBooleanArray("Sensor Statuses", sensorStatus);
+            if (sensorStatus[2]) {
+              int nextOpenSensor = CDSSubsystem.getNextOpenSensor(sensorStatus);
```
azvanderpas: getNextOpenSensor should be rewritten to not use a parameter now that sensorStatus is a instance variable in CDSSubsystem

I also think the front sensor should be omitted from the possible return indices. It should never be an 'open' spot.

src/main/java/frc/robot/Constants.java
```diff
@@ -185,22 +185,23 @@ public String getName() {
   public static final int initialBallSensorChannel = 0;
   public static final int middleBallSensorChannel = 1;
   public static final int finalBallSensorChannel = 2;
+  public static final int sensorsDownLimit = 3;
```
azvanderpas: With #187, this should probably be increased to something between 10 and 20 to cover cases where a cached bad value is returned rather than a re-read.

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -40,79 +40,95 @@ public void initialize() {}
...
 
   // Returns true when the command should end.
   @Override
   public boolean isFinished() {
-    return false;
+    if (CDSSubsystem.getSensorDown() >= Constants.sensorsDownLimit) {
+      return true;
+    } else {
+      return false;
+    }
```
azvanderpas:
```suggestion
    return CDSSubsystem.getSensorDown() >= Constants.sensorsDownLimit;
```

azvanderpas (changes requested):

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
zainmfjavaid: the way the loop works, the front sensor shouldn't be included at all

src/main/java/frc/robot/RobotContainer.java
sarango2158: Unresolved this comment because it still is attached to the AButton on the branch

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -53,21 +58,39 @@ public void execute() {
...
         // A ball count of over 2 OR ball color is wrong and test mode is off (meaning ball color
         // shouldn't be disregarded)
-        // TODO This is somehow ejecting any ball regardless of color
-        if ((lastBallCount > 2)
-            || (sensorStatus[2]
-                && CDSSubsystem.getAllianceColor() != CDSSubsystem.senseColor()
-                && !Constants.testMode)) {
+
+        boolean runEject = false;
+        String reason = "";
```
sarango2158: Future improvement: could make this an enum

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -53,21 +58,39 @@ public void execute() {
...
+          colorEject++;
+        }
+
+        if (runEject) {
           intakeSubsystem.toggleIntake(true);
           CDSSubsystem.CDSWheelToggle(true);
           ejectRunning = true;
+          SmartDashboard.putBoolean("Eject Running", ejectRunning);
+          SmartDashboard.putString("Eject Reason", reason);
+          SmartDashboard.putNumber("Color Eject Tally", colorEject);
```
sarango2158, line 80: May consult with Shuffleboard team how this should fit into their layout

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -40,7 +43,9 @@ public void initialize() {}
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
```
sarango2158, line 47: Do we need to run the cargo stopper motor in reverse for this command too, so we don't overrun it?

src/main/java/frc/robot/commands/CombinedIntakeCDSForwardCommand.java
```diff
@@ -5,22 +5,29 @@
...
 
   public CombinedIntakeCDSForwardCommand(
-      IntakeSubsystem mIntakeSubsystem, CDSSubsystem mCDSSubsystem) {
+      IntakeSubsystem mIntakeSubsystem,
+      CDSSubsystem mCDSSubsystem,
+      ShooterSubsystem mShooterSubsystem) {
     // Use addRequirements() here to declare subsystem dependencies.
     addRequirements(mIntakeSubsystem);
     addRequirements(mCDSSubsystem);
+    addRequirements(mShooterSubsystem);
```
sarango2158, line 27: To test: make sure adding Shooter subsystem as a requirement doesn't prevent us from using the shoot commands when we expect to. Should all work seamlessly together.

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -182,9 +185,21 @@ public String senseColor() {
...
   }
 
-  /*public boolean managementEnabled() {
-    return SmartDashboard.getBoolean("Ball Management Enabled", true);
-  }*/
+  public boolean sensorsOnline() {
+    sensorStatuses = colorSensors.getProximities();
+    for (int prox : sensorStatuses) {
+      if (prox == 0) {
+        sensorsDown += 1;
```
sarango2158, line 193: Should we make this a bool instead? It will not count how many sensors are down because we break the loop on the first one we find. It will however accumulate every time we call it and a sensor is down, even the same sensor as before. Might be easier to use this value if it is true/false

sarango2158 (changes requested):

Ariya3:
not merging with main

PR #189, opened 2022-03-16
Climb base

### LINKED ISSUE #184: Climber needs to run a bit during the match
I don't really have any good mechanical ideas to keep the climber down at this point.
When its winched down it tends to slowly creep up, while the robot is moving about. That can lead to problems if it gets to high and we smack into the low rung.

So I propose we fix the issue in code. 
Lets run the neo at a small percentage in  the down direction during the whole match, lets say like 10% to start? While the neo is running at this small percentage we should have the current limiting on it set to 40 amps.  Then when we climb set it back to 60 amps.
Should probably be a default command for the subsystem, then use the climb enable functionality to schedule a different with toggleWhenPressed. The different command can run the climbEnable() in the periodic.

ModBoyex:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -173,11 +175,11 @@ private void initCommands() {
     if (limelightSubsystem != null && driveBaseSubsystem != null) {
       limelightAlign = new LimelightAlign(limelightSubsystem, driveBaseSubsystem);
     }
-    if (climbSubsystem != null) {
-      climbSubsystem.setDefaultCommand(climbCommand);
-    }
+
     if ((climbSubsystem != null) && (driveBaseSubsystem != null)) {
-      ClimbEnabling = new ClimbEnable(climbSubsystem, driveBaseSubsystem);
+      climbEnabling = new ClimbEnable(climbSubsystem, driveBaseSubsystem);
```
azvanderpas, line 180: This command is no longer used

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -293,4 +295,11 @@ public static DriveBaseSubsystem getDriveBase() {
     }
     return null;
   }
+
+  public ClimbKeepDown getClimb() {
+    if (climbSubsystem != null) {
+      return climbKeepDown;
+    }
+    return null;
```
azvanderpas: Don't think this is used/needed

azvanderpas (changes requested):

src/main/java/frc/robot/Constants.java
```diff
@@ -79,6 +79,8 @@ public AimModes previous() {
     }
   }
 
+  public static final boolean DebugMode = true;
```
sarango2158: Should be false before we merge into main so we can tag it as competition-ready

src/main/java/frc/robot/RobotContainer.java
sarango2158: If we are getting rid of this command we should unmap it from the button as well.

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -148,9 +150,28 @@ public void resetTargetedHeight() {
...
+
   public void climbEnable() {
     climbEnabbled = !climbEnabbled;
     sbClimbEnabbled.setBoolean(climbEnabbled);
+    if (climbEnabbled) {
+      m_climbMotorControllerOne.setSmartCurrentLimit(60);
+      m_climbMotorControllerTwo.setSmartCurrentLimit(60);
+    } else {
+      m_climbMotorControllerOne.setSmartCurrentLimit(10);
+      m_climbMotorControllerTwo.setSmartCurrentLimit(10);
```
sarango2158, line 173: It might be expensive to change the current limit mid-match. Do we need to do this?

src/main/java/frc/robot/RobotContainer.java
ModBoyex: That's the command that slows the robot down when climb is enabled.

src/main/java/frc/robot/RobotContainer.java
ModBoyex: Fixed

src/main/java/frc/robot/Constants.java
ModBoyex: Fixed

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
ModBoyex: Yes, Andrew wants the have the climb run during the match and he wanted us to have a low current limit, but then we need to change it when we are actually climbing.

src/main/java/frc/robot/RobotContainer.java
ModBoyex: Oh, wait I can just move that to the ClimbKeepDown Command.

sarango2158 (approved):

PR #204, opened 2022-03-24
Auton three plus balls

### LINKED ISSUE #199: [6] Add more autonomous paths
When we pair with alliances that have similar autonomous strategies, it would be nice to have more paths that accomplish the same goal (e.g., two-ball that uses a different cargo) so we can switch between them. Will also need a reasonable way to pick which path we're setting up, and instructions for where to place the robot for each path.

tonyn10:
Still testing auton paths at the moment.


src/main/java/frc/robot/Robot.java
```diff
@@ -99,7 +102,7 @@ public void disabledPeriodic() {
   // This autonomous runs the autonomous command selected by your {@link RobotContainer} class. */
   @Override
   public void autonomousInit() {
-    AutonModes.setWaitTime(waitTimeSlider.getDouble(1));
+    AutonModes.setWaitTime(waitTimeSlider.getDouble(0));
```
caseyjbrotherton: does it make sense to use the constant initialWaitTime here? If we change it again, one less place to have to remember to change it.

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
```diff
@@ -70,14 +70,16 @@
...
 
   public DriveBaseSubsystem(Joystick joystick, boolean usingExternal) {
     driveBaseSpeed = 1;
     driverJoystick = joystick;
 
     motorControllers = new MotorController[4];
-    gyro = new AHRS(Port.kMXP);
-    gyro.reset(); // resets the heading of the robot to 0
+    gyro = new AHRS(SPI.Port.kMXP);
+    gyro.enableLogging(true);
```
caseyjbrotherton, line 81: Not sure we want to enable logging during competition.  We could use the  Constants.DebugMode to only turn this on while we are competing.

( It is used for the shuffleboard tabs. )

caseyjbrotherton: What are your thoughts Tony and Calvin?

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
```diff
@@ -227,6 +230,9 @@ public void periodic() {
     // update odometryreset
     double leftPosition = leftEncoder.getPosition();
     double rightPosition = rightEncoder.getPosition();
+
+    sbGyroInfo.setDouble(gyro.getAngle());
```
caseyjbrotherton: We should remove this, with the dial, this is another network call that isn't necessary

src/main/java/frc/robot/Robot.java
```diff
@@ -56,9 +56,12 @@ public void robotInit() {
 
     robotContainer = new RobotContainer();
 
-    chooser.setDefaultOption("Taxi", Constants.Auton.TAXI); // default is taxi mode
+    // TODO: change to correct default
+    chooser.setDefaultOption(
+        "Test Mode", Constants.Auton.TEST); // default is TEST mode for now, change later
```
caseyjbrotherton: Default is two ball in competition branch.  Does it make sense to update our default here to what we think we should do in competition?  Just to reduce competition day changes?

caseyjbrotherton (changes requested):

src/main/java/frc/robot/Robot.java
azvanderpas: Lets not merge into main with TEST as the default.

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -163,84 +168,156 @@ private void initializeRamseteCommands() {
...
         System.out.println("Auton: Test mode selected.");
-        return testCommand;
-      case TAXI:
-        System.out.println("Auton: Taxi mode selected.");
-        return taxiCommand;
+        return driveTestCommand;
+      case PUSHTAXI:
+        System.out.println("Auton: Push Taxi mode selected.");
+        return pushTaxiCommand;
+      case INTAKETAXI:
```
azvanderpas, line 317: What do we mean by push taxi? Is it just taxi without the intake running?

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -163,84 +168,156 @@ private void initializeRamseteCommands() {
...
+              new WaitCommand(initialWaitTime),
+              fiveBallParallel1,
+              fiveBallRamseteCommands[1],
+              new ShooterPressed(shooterSubsystem, limelightSubsystem, cdsSubsystem, false),
+              fiveBallParallel2,
+              fiveBallRamseteCommands[3],
+              new ShooterPressed(shooterSubsystem, limelightSubsystem, cdsSubsystem, false),
+              fiveBallParallel3,
+              fiveBallRamseteCommands[5],
+              new ShooterPressed(shooterSubsystem, limelightSubsystem, cdsSubsystem, false));
```
azvanderpas, line 293: It seems like each path has a similar structure. 

1. wait
2. intake deadline path
3. return to fender path
4. shoot
5. goto 2

A good future update would be to figure out a iterative (or recursive) way to generate these so you don't have to copy paste so much

azvanderpas (dismissed):

src/main/java/frc/robot/Robot.java
tonyn10: That's a great idea. I'll make a constant called defaultInitialWaitTime.

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
tonyn10: I wasn't sure what it did in the first place so I'll try testing the gyro without it.

src/main/java/frc/robot/Robot.java
tonyn10: I will change that. The default auton mode will probably be three ball or four ball depending on what auton mode turns out best after testing is done.

src/main/java/frc/robot/commands/AutonModes.java
tonyn10: It is taxi but the intake is pointed towards the hub so that when it moves out of the tarmac it will push the enemy's ball with the bumper. The intent was so that there would be an auton option if our intake completely wasn't working and we somehow wanted to sabotage the other team.

caseyjbrotherton (approved):

PR #205, opened 2022-03-24
Add method to configure ColorSensorMuxed measurement rate

azvanderpas:


src/main/java/frc/robot/common/hardware/ColorSensorMuxed.java
```diff
@@ -44,6 +62,46 @@ public ColorSensorMuxed(int... ports) {
     lastColorRead = 0;
   }
 
+  public boolean configureMeasurementRates(MeasurementRate rate) {
+
+    ColorSensorMeasurementRate colorRate;
+    ProximitySensorMeasurementRate proxRate;
+    sensorPeriodInSeconds = rate.period;
```
caseyjbrotherton, line 69: Should we move this to the end of the function and only set the sensorPeriodInSeconds if all sensors were successful?
I don't think we would ever set the interval higher, but if so, it becomes more complicated.

caseyjbrotherton (approved):

azvanderpas:
Getting these changes in as part of #208

PR #206, opened 2022-03-24
PR Competition Changes

zbrotherton:


src/main/java/frc/robot/commands/ShooterHeld.java
```diff
@@ -65,9 +64,9 @@ public void execute() {
         }*/
       }
     } else {
-      m_ShooterSubsystem.setCargoBoolean(false);
+      /*m_ShooterSubsystem.setCargoBoolean(false);
```
caseyjbrotherton, line 67: If this is how we are going to roll, let's just remove the code.

If we don't think this is the right way, let's think through what we should do differently here.

caseyjbrotherton (changes requested):

azvanderpas (approved):

src/main/java/frc/robot/commands/ShooterHeld.java
zbrotherton: I want to keep it there just in case

caseyjbrotherton (approved):

PR #208, opened 2022-03-24
CDS state machine

zainmfjavaid:


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -195,14 +195,14 @@ private void configureButtonBindings() {
     // Intake / CDS
     if (outtakeCommand != null) {
       // spits ball out
-      buttons[Constants.RBumper].whileHeld(outtakeCommand);
+      buttons[Constants.AButton].whileHeld(outtakeCommand);
```
azvanderpas: Keep this as RBumper

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -235,24 +259,73 @@ public int getBallCount() {
...
+    switch (state) {
+      case IDLE:
+        nextOpenSensor = -1;
+        msCurrent = 0;
+
+        if (ballCount > 2 || (getMissedColor() && getMissedSensor())) {
+          state = MangementState.EJECT;
+        } 
+        
+         if (ballCount < 3 && currentOpenSensor != -1 && getMissedSensor()) {
```
azvanderpas: We shouldn't need the missed sensor stuff anymore since this will always be running.

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -40,70 +55,42 @@ public void initialize() {}
...
+    switch (state) {
+      case "IDLE":
+        CDSSubsystem.stopCDS();
+        intakeSubsystem.stopIntake();
+        shooterSubsystem.runCargo(0.0);
+        autoEjectRunning.setBoolean(false);
+        autoIntakeRunning.setBoolean(false);
+
+        msCurrent = 0;
+        nextOpenSensor = -1;
```
azvanderpas: These don't get used either

src/main/java/frc/robot/commands/CDSBallManagementCommand.java
```diff
@@ -40,70 +55,42 @@ public void initialize() {}
...
-            sensorIndex = -1;
-            CDSSubsystem.setReady(true);
-          }
-        }
-      }
-    } else {
-      msCurrent += 20;
+    int ballCount = CDSSubsystem.getBallCount();
+
+    sensorStatus = CDSSubsystem.getSensorStatus();
```
azvanderpas: Shouldn't need this in the command anymore. But I think it needs to be added to the CDSSubsystem SM

src/main/java/frc/robot/commands/CombinedIntakeCDSForwardCommand.java
```diff
@@ -37,7 +37,7 @@ public void initialize() {
     CDSSubsystem.CDSBeltToggle(false);
     CDSSubsystem.CDSWheelToggle(false);
     intakeSubsystem.toggleIntake(false);
-    shooterSubsystem.runCargo(Constants.Shooter.cargoReverse);
+    shooterSubsystem.runCargo(Constants.reverseStopperWheelSpeed);
   }
 
   // Called every time the scheduler runs while the command is scheduled.
```
azvanderpas, line 46: I think you want to move the initialize stuff to execute, and then call the ball management's execute from there

something like this pseudo code
```
//Create new CDSBallManagement command object called ballManagementCmd

public void execute() {
  if (CDSSubsystem.getState() == IDLE) {
    // run CDS/intake/stopper wheel
  } else {
    ballManagementCmd.execute();
  }
``` 
I think this will work without having subsystem conflicts since we aren't using the scheduler to run it.

We might want to do something for the Outtake, but maybe not

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -235,24 +246,47 @@ public int getBallCount() {
...
-    CDSTab = Shuffleboard.getTab("CDS Tab");
-    CDSBeltControllerDirection =
-        CDSTab.add("CDS Belt Direction", "Not Running")
-            .withPosition(2, 0)
-            .withWidget(BuiltInWidgets.kToggleSwitch)
-            .getEntry();
-    CDSWheelControllerSpeed = CDSTab.add("CDS Wheel speed", 0).withPosition(3, 0).getEntry();
-    CDSBeltControllerSpeed = CDSTab.add("CDS Belt speed", 0).withPosition(4, 0).getEntry();
+  public void periodic() {
+    int ballCount = getBallCount();
```
azvanderpas: You should be able to just use the instance variable ballCount instead of redefining it here

src/main/java/frc/robot/commands/CombinedIntakeCDSForwardCommand.java
```diff
@@ -26,23 +26,32 @@ public CombinedIntakeCDSForwardCommand(
...
-    intakeSubsystem.toggleIntake(false);
-    shooterSubsystem.runCargo(Constants.Shooter.cargoReverse);
-  }
+  public void initialize() {}
 
   // Called every time the scheduler runs while the command is scheduled.
   @Override
-  public void execute() {}
+  public void execute() {
+    if (CDSSubsystem.getState().toString() == "IDLE") {
```
azvanderpas: Use the enum directly rather that the string version

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -31,15 +36,27 @@
...
 
   private int ballCount = 0;
 
   private int[] sensorStatuses;
+  private boolean[] activationArray = new boolean[3];
+  private String lastBallColor;
+
+  private int currentProxCycle = 0;
+  private int currentColorCycle = 0;
+  private int cycleWait = 1; // read interval for color sensors
```
azvanderpas: We shouldn't need this logic anymore, though it could be useful in the future. Maybe just add a comment that 1 here means it reads every cycle

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -169,56 +196,78 @@ public void stopCDSWheel() {
...
-    if (Constants.DebugMode) {
-      CDSBallCount.setNumber(ballCount);
-    }
 
-    return beamBreakArray;
+      return activationArray;
+    } else {
+      currentProxCycle++;
+      return activationArray;
+    }
```
```suggestion
    } 
    currentProxCycle++;
    return activationArray;

```

Since we always return activationArray, it doesn't need to be part of the if-block

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -169,56 +196,78 @@ public void stopCDSWheel() {
...
-    boolean frontStatus = sensorStatuses[2] > Constants.frontSensorActivation;
-    boolean[] beamBreakArray = {backStatus, middleStatus, frontStatus};
-
-    ballCount = 0;
-    for (boolean status : beamBreakArray) {
-      if (status) {
-        ballCount++;
+  public boolean[] getSensorStatus() {
+    if (currentProxCycle % cycleWait == 0) {
+      currentProxCycle = 1;
```
azvanderpas: If you apply the suggestion below, set it to 0 here like you had before

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -169,56 +196,78 @@ public void stopCDSWheel() {
...
+    }
   }
 
   public int getNextOpenSensor() {
-
-    boolean[] sensorStatus = getSensorStatus();
     // Starts at 0 and ends short of the centering wheel
-    for (int i = 0; i < sensorStatus.length - 1; i++) {
-      if (!sensorStatus[i]) {
+    for (int i = 0; i < activationArray.length - 1; i++) {
```
azvanderpas, line 237: Add a comment about why it is length-1 (its because the front sensor isn't a valid open spot, right?)

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -169,56 +196,78 @@ public void stopCDSWheel() {
...
+        lastBallColor = "Blue";
+        return "Blue";
+      }
     } else {
-      ballColor.setString("Blue");
-      return "Blue";
+      currentColorCycle++;
+      return lastBallColor;
     }
   }
```
azvanderpas: Look at the suggestion for the getSensorStatus and update accordingly (lastBallColor should always equal your return value)

src/main/java/frc/robot/common/hardware/ColorSensorMuxed.java
```diff
@@ -44,6 +64,46 @@ public ColorSensorMuxed(int... ports) {
...
+      default:
+        colorRate = ColorSensorMeasurementRate.kColorRate100ms;
+        proxRate = ProximitySensorMeasurementRate.kProxRate100ms;
+        break;
+    }
+    for (int p : i2cPorts) {
+      if (setI2cPort(p)) {
+        sensors.configureProximitySensor(ProximitySensorResolution.kProxRes11bit, proxRate);
+        sensors.configureColorSensor(
+            ColorSensorResolution.kColorSensorRes20bit, colorRate, GainFactor.kGain3x);
```
azvanderpas: I think our color sensing got worse because the color measurement rate is dependent on the resolution according to the APDS-9151 datasheet. 20bit made the period 400ms which is slower than our default 100ms @ 18bit. I'll fix and merge into this PR.

azvanderpas (changes requested):

azvanderpas (dismissed):

azvanderpas:
I'll approve once the conflicts are resolved

azvanderpas (approved):

PR #209, opened 2022-03-25
Switch to primarily FF control

zbrotherton:


src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -139,7 +147,11 @@ public void windFlywheel(double rpm) {
     } else {
       DTRPM.setDouble(rpm);
       targetRPM = rpm;
-      flywheelPID.setReference(rpm, CANSparkMax.ControlType.kVelocity);
+      flywheelPID.setReference(
+          rpm,
+          CANSparkMax.ControlType.kVelocity,
+          Constants.Shooter.kMaxISlot,
+          flywheelFF.calculate(rpm / 62.0));
```
azvanderpas: Why 62?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -220,6 +232,7 @@ public void prime() {
 
   @Override
   public void periodic() {
+    SmartDashboard.putNumber("ff", flywheelFF.calculate(2600 / 60));
```
azvanderpas: Is this just logging a constant?

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: Put in for testing, it is completely useless I just forgot that I made the change

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: It was 60 to convert to rps, but it seemed to only be exact at 62 when I was on testing bot
However, I just realized that it was because the 2 shooters have completely different characterizations, so I just changed it back to 60 to work with the competition robot.

azvanderpas (approved):

PR #211, opened 2022-03-26
Climb base (With High Climb)

### LINKED ISSUE #153: [1] Create a traversal climb subsystem / mode
Work with CAD/manufacturing on how traversal climb is supposed to work. Determine what hardware and sensors we'll need for that subsystem. Map out what the climber will need to do and create subsystem methods that accomplish it. 

One robot will have this climber and the other one likely will not. We should think about ways to make sure the subsystem does not interfere with the operation of the rest of the robot code if it is not on the current robot.

### LINKED ISSUE #197: [4] Speed up climb driving
Feedback from drive team was that we can increase the speed of the drive base more when climbing.

ModBoyex:


src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -22,237 +21,439 @@
...
+    } else {
+      DhaHeight1.setNumber(HaHeightOne);
+    }
+
+    m_HaTwo.getPIDCtrl().setReference(HaHeightTwo, CANSparkMax.ControlType.kPosition);
+    if (Constants.DebugMode) {
+      sbHaHeightTwo.setNumber(HaHeightTwo);
+    } else {
+      DhaHeight2.setNumber(HaHeightTwo);
+    }
```
azvanderpas: I'm not sure this is needed for the high climb winch. Check with @AndrewIverson

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -22,237 +21,439 @@
...
+      m_McTwo.setSmartCurrentLimit(60);
+      m_HaOne.setSmartCurrentLimit(60);
+      m_HaTwo.setSmartCurrentLimit(60);
     } else {
-      m_climbMotorControllerOne.setSmartCurrentLimit(10);
-      m_climbMotorControllerTwo.setSmartCurrentLimit(10);
+      m_McOne.setSmartCurrentLimit(10);
+      m_McTwo.setSmartCurrentLimit(10);
+      m_HaOne.setSmartCurrentLimit(10);
+      m_HaTwo.setSmartCurrentLimit(10);
```
azvanderpas: making some constants for these numbers would be nice

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -22,237 +21,439 @@
...
-    if (climbEnabbled
-    /** && !m_limitSwitch.get() */
-    ) {
-
-      joystickAxis = -m_climbJoystick.getRawAxis(Constants.leftJoystickY);
-      if (joystickAxis > 0.1 || joystickAxis < -0.1) {
-        if (joystickAxis > 0) {
-          if (climbHeightOne <= Constants.climbHeightMax) {
-            climbHeightOne = climbHeightOne + (joystickAxis / 10 * 8);
+  public void highArms() {
```
azvanderpas, line 219: You might want to think about how to make this the same method as midClimb() just with parameters to select the motorcontroller/joysticks to use. Then if you have to change something you don't have to remember to do it in both places.

src/main/java/frc/robot/commands/ClimbEnable.java
```diff
@@ -28,7 +28,7 @@ public void initialize() {
     m_subsystem.resetTargetedHeight();
 
     if (m_subsystem.getclimbingenable()) {
-      m_drivesubsystem.setDriveBaseSpeed(0.40);
+      m_drivesubsystem.setDriveBaseSpeed(0.6);
```
azvanderpas: How about we make this a constant so its easy to find where to tweek it?

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -22,237 +21,462 @@
...
+      McHeightTwo = sbMcHeightTwo.getDouble(0);
+    }
+
+    sbClimbEnabble.setBoolean(climbEnabble);
+  }
+
+  public void debugPeriodic() {
+    if (DriverStation.isDisabled() && climbEnabble) {
       climbEnable();
     }
```
azvanderpas: This needs to be in periodic as well, right?

src/main/java/frc/robot/commands/ClimbKeepDown.java
```diff
@@ -5,6 +5,7 @@
 package frc.robot.commands;
 
 import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.Constants;
 import frc.robot.subsystems.ClimbSubsystem;
 
 public class ClimbKeepDown extends CommandBase {
```
azvanderpas: ClimbKeepDown seems like it needs a new name since it does more than keep it down

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
ModBoyex: The code is now changed so I think it's fine

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
ModBoyex: Done

src/main/java/frc/robot/commands/ClimbEnable.java
ModBoyex: Done

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
ModBoyex: Yes you are correct and it is now fixed

src/main/java/frc/robot/commands/ClimbKeepDown.java
ModBoyex: I have renamed it to Climb Periodic bc every thing in it need to run periodicaly

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -164,28 +164,27 @@ public void midClimb() {
...
         if (McjoystickAxis < 0) {
-          if (McHeightOne + (McjoystickAxis / 10 * 6) >= 0) {
-            McHeightOne = McHeightOne + (McjoystickAxis / 10 * 6);
+          if (McHeightOne + (McjoystickAxis * -0.5) <= Constants.McHeightMax) {
+            McHeightOne = McHeightOne + (McjoystickAxis * -0.4);
           }
-          if (McHeightTwo + (McjoystickAxis / 10 * 6) >= 0) {
-            McHeightTwo = McHeightTwo + (McjoystickAxis / 10 * 6);
+          if (McHeightTwo + (McjoystickAxis * -0.5) <= Constants.McHeightMax) {
+            McHeightTwo = McHeightTwo + (McjoystickAxis * -0.4);
```
azvanderpas: Make these magic number into constants. Also, we are using only one position value since each side should always match. Remove McHeightTwo

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -195,30 +194,39 @@ public void highArms() {
...
         if (HajoystickAxis < 0) {
-          if (HaHeightOne + (HajoystickAxis / 10 * 6) >= 0) {
-            HaHeightOne = HaHeightOne + (HajoystickAxis / 10 * 6);
+          if (HaHeightOne + (HajoystickAxis / 10 * 6) >= Constants.HaHeightMin) {
+            HaHeightOne = HaHeightOne + (HajoystickAxis * 0.5);
           }
-          if (HaHeightTwo + (HajoystickAxis / 10 * 6) >= 0) {
-            HaHeightTwo = HaHeightTwo + (HajoystickAxis / 10 * 6);
+          if (HaHeightTwo + (HajoystickAxis / 10 * 6) >= Constants.HaHeightMin) {
+            HaHeightTwo = HaHeightTwo + (HajoystickAxis * 0.5);
```
azvanderpas: Same as McHeight comment

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -228,6 +236,21 @@ public void periodic() {
...
+    SmartDashboard.putNumber("Mc2 Hight", m_McTwo.getEncoder().getPosition());
+    SmartDashboard.putNumber("Mc2 IAccum", m_McTwo.getPIDCtrl().getIAccum());
+
+    SmartDashboard.putNumber("Ha1 Applied Output", m_HaOne.getAppliedOutput());
+    SmartDashboard.putNumber("Ha1 Hight", m_HaOne.getEncoder().getPosition());
+    SmartDashboard.putNumber("Ha1 IAccum", m_HaOne.getPIDCtrl().getIAccum());
+
+    SmartDashboard.putNumber("Ha2 Applied Output", m_HaTwo.getAppliedOutput());
+    SmartDashboard.putNumber("Ha2 Hight", m_HaTwo.getEncoder().getPosition());
+    SmartDashboard.putNumber("Ha2 IAccum", m_HaTwo.getPIDCtrl().getIAccum());
```
azvanderpas: Move to debugPeriodic

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
ModBoyex: For a leter date

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -135,9 +135,11 @@ public void climbKeepDownFunction() {
 
     m_HaOne.getPIDCtrl().setReference(HaHeightOne, CANSparkMax.ControlType.kPosition);
     sbHaHeightOne.setNumber(HaHeightOne);
+    m_HaOne.getPIDCtrl().setIMaxAccum(0.25, 0);
 
     m_HaTwo.getPIDCtrl().setReference(HaHeightTwo, CANSparkMax.ControlType.kPosition);
     sbHaHeightTwo.setNumber(HaHeightTwo);
+    m_HaTwo.getPIDCtrl().setIMaxAccum(0.25, 0);
```
azvanderpas: MaKe A CoNsTaNt

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -162,21 +164,23 @@ public boolean getclimbingenable() {
...
-            McHeightTwo = McHeightTwo + (McjoystickAxis * -1.5);
+            McHeightTwo = McHeightTwo + (McjoystickAxis * Constants.McUpSpeed);
           }
         }
         if (McjoystickAxis < 0) {
           if (McHeightOne + (McjoystickAxis * -0.5) <= Constants.McHeightMax) {
-            McHeightOne = McHeightOne + (McjoystickAxis * -0.4);
+            McHeightOne = McHeightOne + (McjoystickAxis * Constants.McDownSpeed);
           }
           if (McHeightTwo + (McjoystickAxis * -0.5) <= Constants.McHeightMax) {
```
azvanderpas: Use your constant in the if conditions too

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -191,21 +195,22 @@ public void midClimb() {
...
           }
         }
         if (HajoystickAxis < 0) {
           if (HaHeightOne + (HajoystickAxis / 10 * 6) >= Constants.HaHeightMin) {
-            HaHeightOne = HaHeightOne + (HajoystickAxis * 0.5);
+            HaHeightOne = HaHeightOne + (HajoystickAxis * Constants.HaSpeed);
           }
           if (HaHeightTwo + (HajoystickAxis / 10 * 6) >= Constants.HaHeightMin) {
-            HaHeightTwo = HaHeightTwo + (HajoystickAxis * 0.5);
+            HaHeightTwo = HaHeightTwo + (HajoystickAxis * Constants.HaSpeed);
```
azvanderpas: Conditions need to match constants as well

azvanderpas (changes requested):

azvanderpas (dismissed):

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -183,6 +188,9 @@ private void initializeRamseteCommands() {
               ramsetes[i],
               new CombinedIntakeCDSForwardCommand(intakeSubsystem, cdsSubsystem, shooterSubsystem));
     }
+    parallels[0] =
+        parallels[0].deadlineWith(new InstantCommand(climbSubsystem::deployHA, climbSubsystem));
+
```
azvanderpas: After thinking about this, this will break the ramsete command because the deployHA will finish first and cause the others in the deadline to stop.

azvanderpas:
Need tony approval on auton changes

azvanderpas (approved):

PR #217, opened 2022-03-29
Finetuned auton

tonyn10:


src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -76,7 +76,7 @@ public void end(boolean interrupted) {
   // Returns true when the command should end.
   @Override
   public boolean isFinished() {
-    if (i >= 15) { // 1000 miliseconds delay TODO: Use a CDS method for this when possible
+    if (i >= 20) { // 1000 miliseconds delay TODO: Use a CDS method for this when possible
```
azvanderpas: Update comment to match new value (20 ms * 20 = 400 ms)

azvanderpas (changes requested):

src/main/java/frc/robot/commands/ShooterPressed.java
azvanderpas: CDSSubsystem getBallCount() function should be able to be used for an end condition check. But do that in a new branch/PR after testing

azvanderpas (approved):

Issue #216, opened 2022-03-29
CDS eject from shooter


azvanderpas:
Sometimes cycling wrong colored balls out through the shooter would be better to get them out of our way.

Need to discuss with drive team to determine exactly what cases we want to do this in. 

Initial case thoughts:
- intake wrong color with no other cargo in robot
- wrong color makes it to middle or top sensors (eject until only correct color remains)
- manually with operator button press (?)
- instead of ejecting correct color 3rd ball out front (?)

azvanderpas:
I think a good way to tackle this is to refactor our state machine to use all the sensors as inputs rather than mostly the front/bottom. Here is a enumeration of the different ways the sensors could read colors. If we use this table to determine what the CDS should be doing, we may not even need to use proximity and instead just rely on the colors. 

The bottom/front sensor may be tricky since it's color sense may be noisier. Might work better to use proximity on the front and colors on the middle/top. 

This becomes less of a state machine with this setup, since the the case is determined by the measurements only and not the previous state. 

0 = empty
1 = wrong color
2 = correct color

Bottom|Middle|Top
000 - empty/idle
001 - shooter eject
002 - idle
010 - shooter eject
011 - shooter eject
012 - front eject
020 - advance
021 - shooter eject
022 - idle
100 - shooter eject
101 - shooter eject
102 - front eject
110 - shooter eject
111 - shooter eject
112 - front eject
120 - front eject 
121 - shooter and front eject
122 - front eject
200 - advance
201 - shooter eject
202 - advance
210 - shooter eject
211 - shooter eject
212 - shooter eject (or front eject)
220 - advance
221 - shooter eject
222 - shooter eject (or front eject)

IDLE
000 - empty/idle
002 - idle
022 - idle

SHOOTER_EJECT
001 - shooter eject
010 - shooter eject
011 - shooter eject
021 - shooter eject
100 - shooter eject
101 - shooter eject
110 - shooter eject
111 - shooter eject
201 - shooter eject
210 - shooter eject
211 - shooter eject
212 - shooter eject (or front eject)
221 - shooter eject
222 - shooter eject (or front eject)

ADVANCE
020 - advance
200 - advance
220 - advance
202 - advance

FRONT_EJECT
012 - front eject
102 - front eject
120 - front eject


CORNER CASE
121 - shooter and front eject

azvanderpas:
[...]

PR #218, opened 2022-03-30
CDS state machine

zainmfjavaid:


azvanderpas:
I think we are going to go back to the drawing board a bit with #216, so I'm going to close this review for now.

PR #220, opened 2022-03-31
Auton three plus balls

azvanderpas:


src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -45,7 +46,7 @@ public void initialize() {
   @Override
   public void execute() {
     m_ShooterSubsystem.prime();
-    if (m_ShooterSubsystem.wheelReady()) {
+    if (m_ShooterSubsystem.wheelReady() || i > 0) {
```
azvanderpas, line 48: If we still need this it wasn't the battery

src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -55,6 +56,7 @@ public void execute() {
           m_ShooterSubsystem.setCargoBoolean(true);
         }
         i++;
+        SmartDashboard.putNumber("ShooterPressed i count", i);
```
caseyjbrotherton: Do we want this removed?

caseyjbrotherton (dismissed):

tonyn10 (approved):

PR #221, opened 2022-04-03
Auton code reorganized

tonyn10:
More memory efficient and easier to read


azvanderpas:
Not sure if we'll have time to test these changes before state champs. This is going to be low priority since Auton is fairly stable and consistent.

tonyn10:
We want a better two ball auto (have the option to grab and shoot the enemy's balls away from them) so I think this code will be relevant.

caseyjbrotherton (dismissed):
Looks sweet.

Issue #224, opened 2022-04-04
Better shooter accuracy


azvanderpas:
Feedforward control: https://docs.wpilib.org/en/stable/docs/software/advanced-controls/controllers/feedforward.html#using-feedforward-to-control-mechanisms

Feedforward + PID control: https://docs.wpilib.org/en/stable/docs/software/advanced-controls/controllers/combining-feedforward-feedback.html

zbrotherton:
So we should switch to WPIlib PID to use both in setVoltage? 
I thought sparkMax's setReference included the arbFF param as a way to get around using WPIlib PID if we wanted to use FF control, but I could be wrong

azvanderpas:
Maybe you are right. Its odd that I don't see any examples of that being used anywhere though.

Either way we need to fix the shooter tuning since it appears we aren't staying stable, or at least we aren't checking that we are. Possibly the setReference doesn't do voltage compensation correctly since it's affected by our belt speed?

zbrotherton:
It is really odd, I was looking for a long time for any sort of documentation or examples, and there was nothing.
I'll work with keeping PID stable, though

caseyjbrotherton:
Talks from whiteboard:
  - Need to characterize when we have 2 wheels from manufacturing. ( RPM may need to be higher because of less spin, and less magic forces. )
  -  Need to check and see that RPM holds after sending cargo.  Right now, we use an exponential smoothing, but once we hit that point, we send all balls without rechecking RPM. 
  - Investigate Auton and belt issue.  Worked around it by just sending the ball even if the RPM doesn't hit.  This seemed to work during competition, however, it is a mystery and we should tune belt speed, or something to make sure that auton can launch the ball.

PR #227, opened 2022-04-04
Automatic climb project

### LINKED ISSUE #225: Automate Climb
So I think it should be possible to fully automate climbing(besides lining up) in two commands, first a pressed to release the climber, and a while held to climb so we can abort if needed and resume to manual control.

- use velocity based pid to unwinch to the point that the spool can go out at full speed with out tangling
- use position pid to unwinch until full extension setpoint

Second command

- Pull up with velocity pid until close to position
- Move poles as required(we can use a gyro to tell if we are tilted and move the poles. Or could just assume they need to be in a certain spot at certain main winch positions
- Finish pull up with position command
- Once at the required height we move the poles using positional pid to shift the robot back(until the poles are at the correct position(maintaining the winch position))
- Lower the robot slightly by adjusting the setpoint for the main winches
- Passive hooks should be above the traversal rung at this point
- Then we move our poles to a disengaged set point and release the winch.

The steps can be seen executing in this gif
https://imgur.com/a/Nxa1mFn (edited)

### LINKED ISSUE #228: Latch to prevent climber from deploying early
http://wiki.team2537.com/wiki/ControlSystems/SoftwareTeam/Training/GettingStarted/ServoMotor

We have a servo motor that we will be using to hold the arms to prevent them from deploying early.

This will need to go into enable climb.

zbrotherton:


zbrotherton:
Closes #225

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
```diff
@@ -57,7 +57,7 @@
   private SimpleMotorFeedforward simpleMotorFF;
 
   private final DifferentialDriveOdometry odometry;
-  private AHRS gyro;
+  public static AHRS gyro;
```
caseyjbrotherton, line 60: I don't understand how this fits with the changes, or why it's needed at all.  Doesn't seem like gyro is referenced outside of drivebasesybsystem.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -165,6 +165,10 @@ public void runCargo(double speed) {
     stopperController.set(speed);
   }
 
+  public boolean wheelReady(int low, int high) {
+    return (smoothRPM > targetRPM - low && smoothRPM < targetRPM + high);
+  }
+
```
caseyjbrotherton: This doesn't look used at all, ,either.

src/main/java/frc/robot/commands/ShooterPressed.java
```diff
@@ -50,7 +50,7 @@ public void execute() {
       // Otherwise, alignment is checked.
       if (i > 0 || !LLEnabled || m_LimelightSubsystem.calculatePID() == 0.0) {
         if (i == 0) {
-          m_CDSSubsystem.CDSBeltToggle(false);
+          m_CDSSubsystem.CDSBeltToggle(false, 0.95);
```
caseyjbrotherton: same comment as in other PR.  You have a constant for this

src/main/java/frc/robot/Constants.java
```diff
@@ -27,7 +27,8 @@ public AimModes previous() {
     },
     EJECT(2500.0, 0.0),
     LAUNCH(0.0, 0.0),
-    TARMAC(2530.0, 0.0),
+    TARMAC(2480.0, 0.0),
+    ATARMAC(2480.0, 0.0),
```
caseyjbrotherton: Same comment.  These are the same, wonder if they should be different.  Is one adjusted in case of over-inflated balls?

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
zbrotherton: The gyro is necessary for some sequences of the automatic climb.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
zbrotherton: It is not used, it was just added at competition in case we wanted to narrow or widen our deadband for a certain command.

src/main/java/frc/robot/Constants.java
zbrotherton: Answered in other PR- 

"We added atarmac and tarmac as a way to change the auto ball speed and tele ball speed independently, as they are shooting differently somehow. This is part of a larger issue tracked as https://github.com/AusTINCANsProgrammingTeam/2022RobotCode/issues/224"

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -245,20 +252,20 @@ public void initAuton(Constants.Auton mode) {
...
       case PUSHTAXI: // the two taxis
       case INTAKETAXI:
-        if (driveBaseSubsystem != null && intakeSubsystem != null && cdsSubsystem != null) {
+        /* if (driveBaseSubsystem != null && intakeSubsystem != null && cdsSubsystem != null) {
           autonModes = new AutonModes(driveBaseSubsystem, intakeSubsystem, cdsSubsystem);
         } else {
           success = false;
         }
-        break;
+        break;*/
```
caseyjbrotherton: I'm assuming that the code was commented out to try and optimize the run time?

The cycles saved are minimal if any.    Don't need to change it, but in this case, I would have optimized for clarity instead of speed.

zbrotherton: It was commented in case we needed it, but we don't so I'm removing it entirely

src/main/java/frc/robot/subsystems/DriveBaseSubsystem.java
caseyjbrotherton: ahh.  So, will be part 2, if that happens.  k

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -103,409 +106,276 @@
...
-    SmartDashboard.putNumber("Ha2 Hight", m_HaTwo.getEncoder().getPosition());
-    SmartDashboard.putNumber("Ha2 IAccum", m_HaTwo.getPIDCtrl().getIAccum());
-
-    // Mid Climb
-    if (sbMcHeightOne.getDouble(0) != McHeightOne) {
-      McHeightOne = sbMcHeightOne.getDouble(0);
+  // functions for ClimbSequence1
+  public void deployArms() {
+    if (armEncoderHeightOne < Constants.armHeightFeather
+        && armEncoderHeightTwo < Constants.armHeightFeather) {
```
caseyjbrotherton: OR?

Problem you are trying to automate is that the initial movement of the arms needs to be slow.  If one of the arms is below that minimum safe point, you should probably continue using the velocity PID.

zbrotherton: true, must have slipped past

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -103,409 +106,276 @@
...
-  @Deprecated
-  public void runManual() {
-    if (climbEnabble) {
+  public boolean atFirstSetpoint() {
+    // checking if the armHeight is within an acceptable range
+    return (armEncoderHeightOne > Constants.armHeightMin - Constants.climbArmDeadband
+            && armEncoderHeightOne < Constants.armHeightMin + Constants.climbArmDeadband)
+        && (armEncoderHeightTwo > Constants.armHeightMin - Constants.climbArmDeadband
+            && armEncoderHeightTwo < Constants.climbArmDeadband + Constants.climbArmDeadband);
+  }
```
caseyjbrotherton, line 389: This is the end condition of the automatic deploy.

It feels like you could get caught in an edge case with each arm trying to get within the correct range.
It might not be a big problem with the only opposing forces being the gas springs.  

Should check during testing and see if the position for either arm oscillates around your armHeightMin value, and if so, do something here to let it track each arm independently.

src/main/java/frc/robot/commands/ClimbSequence1.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public void execute() {
+    climbSubsystem.deployArms();
+    climbSubsystem.resetClimbHeights();
+  }
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {
+    climbSubsystem.resetClimbHeights();
```
caseyjbrotherton, line 44: Notify the operator tab?  So that you know when auto deploy is done?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -223,7 +226,11 @@ private void configureButtonBindings() {
     }
 
     if (climbSubsystem != null) {
-      buttons2[Constants.startButton].whenPressed(climbEnabling);
+      // enable climb and spool out arms
+      buttons2[Constants.startButton].whenPressed(
+          new SequentialCommandGroup(climbEnabling, climbSequence1));
```
caseyjbrotherton: Consider a timeout here?

In case there is something wrong with spooling out the arms, so that you can attempt to do something manually?

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
zbrotherton: Got it, I'll watch how this acts during testing.

src/main/java/frc/robot/RobotContainer.java
zbrotherton: How could I easily add a timeout for this? Would I just do something similar to my increment variable for shooter pressed?

zbrotherton: Or maybe I could check for a network table button as well?

caseyjbrotherton: https://docs.wpilib.org/en/stable/docs/software/commandbased/convenience-features.html?highlight=timeout#withtimeout

I don't know if that works.  but description seems reasonable.

You could also have another button call cancel

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -17,495 +18,321 @@
...
-      sbMcTargettedOne.setDouble(McHeightOne);
-    }
-    sbMcSpeedOne.setDouble(m_McOne.getEncoder().getVelocity());
-
-    if (sbMcHeightTwo.getDouble(0) != McHeightTwo) {
-      McHeightTwo = sbMcHeightTwo.getDouble(0);
+  // functions for ClimbSequence1
+  public void deployArms() {
+    if (armEncoderHeightOne > Constants.armHeightFeather
+        || armEncoderHeightTwo < Constants.armHeightFeather) {
```
azvanderpas: Not sure which, but these should both be > or <

src/main/java/frc/robot/commands/ClimbSequence1.java
```diff
@@ -23,14 +24,17 @@ public ClimbSequence1(ClimbSubsystem s) {
...
   }
 
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
+    if(i>5){
     climbSubsystem.deployArms();
+    }
     climbSubsystem.resetClimbHeights();
+    i++;
```
azvanderpas, line 37: Unbounded increment isn't super great, but it won't overflow for many hours so it's fine for now.

azvanderpas (dismissed):

azvanderpas (approved):

PR #229, opened 2022-04-05
Intake deploy

caseyjbrotherton:
Opening a draft pull request to review this code before the meeting tonight


src/main/java/frc/robot/Constants.java
```diff
@@ -178,8 +178,11 @@ public String getName() {
   public static final boolean usingExternal = false;
   public static final int encoderCountsPerRev = 8192;
 
-  // Intake Contstants
+  // Intake Constants
   public static final int intakeMotorOneID = 1;
+  public static final int intakeDeployMotorID = 24;
```
caseyjbrotherton: This will need to be fixed.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -204,6 +207,9 @@ private void configureButtonBindings() {
     } /*else {
         buttons[Constants.RTriggerButton].whileHeld(intakeForwardCommand);
       }*/
+    if (intakeDeployCommand != null) {
+      buttons2[Constants.RBumper].whenPressed(intakeDeployCommand);
```
caseyjbrotherton: Rbumper is already used to run outtake.  Need a better place for this.

src/main/java/frc/robot/commands/IntakeDeploy.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {}
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    return true;
```
caseyjbrotherton, line 36: This will run only once with isFinished set to true..

Is this what we want to happen?  One press equals a swap in state of the intake?
We could also consider whether we retract the intake while a button is held.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -54,6 +97,15 @@ public void toggleIntake(boolean reverse) {
     }
   }
 
+  public void deployIntake() {
```
caseyjbrotherton, line 82: This is set up as a toggle.  

Not sure we want this.  Feel like we should have deployIntake and retractIntake, with a third method that returns a boolean when the intake is deployed.  

This way, auton can have deployIntake, and be certain that it is not going to toggle anything, but will absolutely deploy.

If we change the button from a whenpressed to a whileheld, we can specify exactly what happens.  retract upon initialize, nothing during execute, and deploy at end.

And if you end up needing a toggle, the code is still easy:

if ( intakeDeployed() )
  retractIntake()
else
  deployIntake()

src/main/java/frc/robot/Constants.java
Asher123456789: Done

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -122,10 +122,11 @@ public void periodic() {
...
     }
     sbintakeDeployed.setBoolean(intakeDeployed);
-    sbintakeDeployPosition.setDouble(deployEncoder.getPosition());
+    //sbintakeDeployPosition.setDouble(deployController.getEncoder().getPosition());
 
-    deployController.updateSmartDashboard();
+    //deployController.updateSmartDashboard();
     SmartDashboard.putBoolean("Intake Out?", intakeDeployed);
-    SmartDashboard.putNumber("Deploy Encoder", deployEncoder.getPosition());
+    //System.out.println("encoder: " + deployController.getEncoder().getPosition());
```
caseyjbrotherton: These are my fault, will clean them up before we merge

src/main/java/frc/robot/RobotContainer.java
Asher123456789: Deleted completely in favor of an automatic intake deploy and retract system.

ModBoyex:
10/10

azvanderpas (approved):

PR #230, opened 2022-04-05
CDS Shooter Eject/Simulation

zainmfjavaid:


src/main/java/frc/robot/Constants.java
```diff
@@ -199,6 +199,12 @@ public String getName() {
   public static final int frontSensorActivation = 200;
   public static final int middleSensorActivation = 450;
   public static final int backSensorActivation = 600;
+  public static final String[] idleStates = {"000", "002", "022"};
+  public static final String[] advanceStates = {"020", "200", "220", "202"};
+  public static final String[] intakeEjectStates = {"012", "102", "120"};
+  public static final String[] shooterEjectStates = {
+    "001", "010", "011", "021", "100", "101", "110", "111", "201", "210", "211", "212", "221", "222"
+  };
```
azvanderpas: Should be 27 different combinations. I only count 24

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -245,27 +279,71 @@ public int getNextOpenSensor() {
...
     if (currentColorCycle % cycleWait == 0) {
       currentColorCycle = 0;
-      Color[] colors = colorSensors.getColors();
-      SmartDashboard.putNumber("Front Sense B", colors[2].blue);
-      SmartDashboard.putNumber("Front Sense R", colors[2].red);
+      colors = colorSensors.getColors();
+
+      double magnitude = colors[2].red + colors[2].green + colors[2].blue;
+      double redRatio = colors[2].red / magnitude;
+      double blueRatio = colors[2].blue / magnitude;
```
azvanderpas: This turns out to be redundant. ColorSensorV3 already scales this way. So the magnitude here will always be 1

src/main/java/frc/robot/subsystems/CDSSubsystem.java
```diff
@@ -302,67 +380,66 @@ public ManagementState getState() {
...
+        ballLayoutArray[i] = "1";
+      }
+    }
 
-    switch (state) {
-      case IDLE:
-        nextOpenSensor = -1;
-        msCurrent = 0;
+    ballLayout =
+        String.format("%s%s%s", ballLayoutArray[0], ballLayoutArray[1], ballLayoutArray[2]);
```
```suggestion
    ballLayout = ballLayoutArray[0] + ballLayoutArray[1] + ballLayoutArray[2];
```

Also, make sure the order is correct. I thought index 2 was the front sensor?

azvanderpas (changes requested):

PR #232, opened 2022-04-05
Auton better two ball

tonyn10:
Has changes from AutonRemodel branch. Has a meannn two ball auto.


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -136,7 +133,7 @@ private void initSubsystems() {
 
     limelightSubsystem = new LimelightSubsystem();
 
-    climbSubsystem = new ClimbSubsystem(operatorJoystick);
+    // climbSubsystem = new ClimbSubsystem(operatorJoystick);
```
azvanderpas: Can't merge with this still commented

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -149,38 +101,12 @@ public AutonModes(
       } else {
         ramseteCommands[i] = r;
       }
-      ramseteCommands[i] =
-          ramseteCommands[i].andThen(
-              () -> driveBaseSubsystem.stopDriveMotors()); // stop drive motors after each ramsete
+      // stopDriveMotors might just be an unnesessary command to the sequence
```
azvanderpas, line 106: Make sure to confirm this

.SysId/sysid.json
```diff
@@ -1,6 +1,6 @@
 {
   "SysId": {
-    "Analysis Type": "Drivetrain",
+    "Analysis Type": "General Mechanism",
```
azvanderpas: Is this right?

src/main/java/frc/robot/commands/AutonModes.java
```diff
@@ -189,135 +115,130 @@ private void initializeRamseteCommands() {
               ramsetes[i],
               new CombinedIntakeCDSForwardCommand(intakeSubsystem, cdsSubsystem, shooterSubsystem));
     }
-    parallels[0] =
-        parallels[0].alongWith(new InstantCommand(climbSubsystem::deployHA, climbSubsystem));
+    // parallels[0] =
+    //     parallels[0].alongWith(new InstantCommand(climbSubsystem::deployHA, climbSubsystem));
```
azvanderpas: Can't merge without uncommenting

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -230,50 +227,53 @@ private void configureButtonBindings() {
...
+
+    AutonModes autonMode =
+        new AutonModes(
+            mode,
+            driveBaseSubsystem,
+            shooterSubsystem,
+            limelightSubsystem,
+            cdsSubsystem,
+            intakeSubsystem,
+            null);
```
azvanderpas: Can't merge with this null

azvanderpas (changes requested):

.SysId/sysid.json
tonyn10: I believe that's the sysid for shooter

src/main/java/frc/robot/commands/AutonModes.java
tonyn10: Two ball works without it. I assume it will work the same for three ball and four ball but I haven't gotten the chance to test.

src/main/java/frc/robot/Constants.java
```diff
@@ -218,7 +219,7 @@ public String getName() {
   public static final int sensorsDownLimit = 3;
 
   // CDS Constants
-  public static final boolean ballManagementEnabled = true;
+  public static final boolean ballManagementEnabled = false;
```
azvanderpas: I think we need to put this back

src/main/java/frc/robot/Constants.java
```diff
@@ -25,7 +25,7 @@ public AimModes previous() {
         return values()[values().length - 1];
       }
     },
-    EJECT(2500.0, 0.0),
+    EJECT(1500.0, 0.0),
```
azvanderpas, line 28: This is going to be way too high with the new shooter tuning

azvanderpas (changes requested):

src/main/java/frc/robot/Constants.java
tonyn10: This aim mode isn't going to be used anyways unless cds/intake is using it. The mean two ball uses the Outtake command instead.

azvanderpas (dismissed):

.Glass/glass.json
```diff
@@ -0,0 +1,143 @@
+{
```
azvanderpas, line 1: Does this file need to be tracked in git?

azvanderpas (changes requested):
Spotless needs to get run (just build while connected to the internet and commit/push the whitespace changes)

.Glass/glass.json
AndrewIverson: If it is then glass will open with the same settings everytime. Glass is like a better shuffleboard that doesn't suck at graphing. I don't really care if the file gets tracked but yall should use glass anytime you need to graph something its such a better user experience.

azvanderpas (approved):

PR #234, opened 2022-08-31
Battery Checker

calvinwt:
Simple program to check the voltage and input current of the battery, and then print it to shuffleboard. Also has the capability to run a timer when the robot first starts, changing a boolean when it ends.


src/main/java/frc/robot/Constants.java
```diff
@@ -16,6 +16,8 @@
  */
 public final class Constants {
 
+  public static final int minvoltage = 10;
```
caseyjbrotherton: Feel like this should start higher.  at 10.5 mininum, or 11.

from what I am seeing on ChiefDelphi, 10.5V is seen as the point where battery performance drops quickly to zero.
And we don't want the batteries to be harmed by draining them. 

 ( And because we can't shut down the robot, we should assume that there will be some current drop while a team member walks onto the field to switch the battery. )

https://www.chiefdelphi.com/t/about-battery-life/126825
https://www.chiefdelphi.com/t/cba-5-and-battery-testing/408305

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,78 @@
...
+
+  private ShuffleboardTab btTab;
+  private NetworkTableEntry sbVoltage;
+  private NetworkTableEntry sbInCurrent;
+  private NetworkTableEntry sbTimer;
+  private NetworkTableEntry sbTimerChange;
+  private Timer timer;
+
+  private void init() {
+    timer.reset();
```
caseyjbrotherton: timer = new Timer();

This code only counts the time since the robot code reset the timer.  Can we get since the robot was turned on?

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,78 @@
...
+    timer.start();
+  }
+
+  public BatterySubsystem() {
+    init();
+  }
+
+  public void periodic() {
+    sbVoltage.setDouble(getVoltage());
+    sbInCurrent.setDouble(getInputCurrent());
```
caseyjbrotherton: We would like to track high current usage.  Can we accumulate times of high current usage?
It would be ok to count "ticks"  approximately 20ms intervals...
But it would be even better to accumulate the real time in between ticks.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,78 @@
...
+    sbTimer.setDouble(getTimer());
+    sbTimerChange.setBoolean(
```
caseyjbrotherton: We want to warn if any of the three conditions happen:
1.  High current usage for some amount of time
2. Any usage for a longer time
3. Dangerously low voltage. 

It would also be great to have a "yellow zone" of optional alert, and then have a more forceful, "danger zone" alert.
Maybe even disable the robot in the danger zone?

Maybe yellow zone is 1 or 2?  and danger zone is passing 3?  Not sure what is best.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,78 @@
...
+  public double getInputCurrent() {
+    return RobotController.getInputCurrent();
+  }
+
+  public double getTimer() {
+    return timer.get();
+  }
+
+  public boolean checkTimer() {
+    if (timer.hasElapsed(300)) {
```
caseyjbrotherton: Let's define 300 as a constant.

src/main/java/frc/robot/commands/BatteryCommand.java
```diff
@@ -0,0 +1,34 @@
...
+    addRequirements(s);
+    subsystem = s;
+  }
+
+  @Override
+  public void initialize() {}
+
+  @Override
+  public void execute() {
+    subsystem.periodic();
```
caseyjbrotherton: subsystem.periodic s run once per run of the command scheduler already?

https://docs.wpilib.org/en/stable/docs/software/commandbased/subsystems.html

Don't think we need this command.

caseyjbrotherton (dismissed):
Good start.  I noted some places that need some tweaks and more thought.

Don't think that we need the default command, but might need to create other commands.

There is another pattern besides what we used last season.  We will be talking about this further in our meetings.  Here is a pre-release document from WPIlib talking about the other pattern:
https://github.com/wpilibsuite/frc-docs/blob/bc87e1e906e608c6517399e5cb6fa45c11eabd42/source/docs/software/commandbased/organizing-command-based.rst

If we need another command, instant commands and instant command factories might be a really good pattern for the Battery Checker.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
caseyjbrotherton: I should note, not sure that one can disable the robot programatically.  But I feel like we need to try and consider a way to make it impossible to ignore, if we can

Is there something more annoying that we can do in the danger zone?  Continuous beeping on the robot?

src/main/java/frc/robot/subsystems/BatterySubsystem.java
azvanderpas: Shuffleboard boolean is good, but you might think about also using [DriverStation.reportError](https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/wpilibj/DriverStation.html#reportError(java.lang.String,boolean))

src/main/java/frc/robot/Constants.java
calvinwt: Completed, changed it to 12 for good measure

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: Completed, moved the reset to robotInit()

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: Completed

src/main/java/frc/robot/commands/BatteryCommand.java
calvinwt: Completed, removed it

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: Completed

calvinwt: Completed, made it delay the timer if the current is below a certain value (still need to find that value though)

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+    return false;
+  }
+
+  public void checkVoltage() {
+    if (getVoltage() < Constants.minVoltageRed) {
+      DriverStation.reportError("Change the Battery!", true);
+    } else if (getVoltage() < Constants.minVoltageYellow) {
+      DriverStation.reportError("Change the Battery Soon!", true);
+    }
```
azvanderpas: Should the "Change the Battery Soon!" cases be reportWarning instead of reportError? 

Also, I think printing the stack trace is unnecessary (boolean parameter of reportError/Warning) and probably will hide the actual message from the operator. 

Also also, once you hit an error case is the error going to reprint every 20 ms forever?

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,120 @@
...
+  public double getVoltage() {
+    return RobotController.getBatteryVoltage();
+  }
+
+  public double getInputCurrent() {
+    return RobotController.getInputCurrent();
+  }
+  public void checkCurrent() {
+    if(getInputCurrent() < Constants.maxBatteryCurrent) {
+      Timer.delay(0.02); //Was current timer, but I got a warning about it needing to be static so I changed it to this. Will this cause conflicts?
```
azvanderpas: We should let the scheduler handle delays and not use this. You can be certain that every call of periodic() happens 20 ms after the last. 

Plus I don't think this delay helps anything. I think the intent was to count the times we read a high current (or start/stop currentTimer to be somewhat more accurate), and then if the battery has experience high current draw for too long since being changed display the error/warning.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+
+  public double getTimer() {
+    return Timer.getFPGATimestamp();
```
azvanderpas: This gets the system clock time, and isn't specific to the Timer's you've created. What you want to use is 
timer.get() and currentTimer.get() to get the times from those specific timers.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,120 @@
...
+    sbInCurrent = btTab.add("Input Current", 0).withSize(2, 2).withPosition(2, 0).getEntry();
+    sbTimer = btTab.add("Timer", 0).withSize(2, 2).withPosition(0, 2).getEntry();
+    sbTimerHighCurrent = btTab.add("High Current Timer", 0).withSize(2, 2).withPosition(0, 4).getEntry();
+    sbTimerChange = btTab.add("Change Timer", 0).withSize(2, 2).withPosition(4, 0).getEntry();
+    timer.start();
+    currentTimer.start();
+  }
+
+  public BatterySubsystem() {
+    init();
```
azvanderpas: Is there a reason we need a separate private method for the constructor?

azvanderpas (changes requested):
I think before anything else we should try to setup simulation and/or unit tests for this. Seems like a good project to learn those skills on.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,120 @@
...
+      DriverStation.reportError("Change the Battery!", true);
+      return true;
+    } else if (timer.hasElapsed(Constants.timeInSecondsHighCurrentYellow)) {
+      DriverStation.reportError("Change the Battery Soon!", true);
+      return false;
+    } else if (timer.hasElapsed(Constants.timeInSecondsGeneralYellow)) {
+      DriverStation.reportError("Change the Battery Soon!", true);
+      return false;
+    }
+    return false;
```
azvanderpas: Each time we redeploy code these timers will be reset. This may not true of Timer.getFPGATimestamp(), but that is somewhat unclear. Either way tracking time elapsed across redeploys presents a problem.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: Completed, yellow and red zone exist only in the actual code, the simulation only uses 1 marker for checking times

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: Done, switched to start() and stop() functions

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: completed

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: fixed to just initialize when called

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: TBD

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: TBD

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: completed, moved to warning and stopped stack trace from printing

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,221 @@
...
+    } else {
+      return RobotController.getBatteryVoltage();
+    }
+  }
+
+  public double getInputCurrent() {
+    return RobotController.getInputCurrent();
+  }
+
+  public void checkCurrent() {
```
azvanderpas, line 122: This method does nothing...

src/main/java/frc/robot/subsystems/BatterySimulation.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    //Forcefully sets the voltage when enabling the robot.
+    if (DriverStationSim.getAutonomous() == true || DriverStationSim.getEnabled() == true) {
+      batterySubsystem.startHCTimer();
+      //RoboRioSim.setVInVoltage(12.1);
+    } else {
+      //RoboRioSim.setVInVoltage(14.0);
+      batterySubsystem.stopHCTimer();
+    }
```
azvanderpas: The timers need to be controlled within the BatterySubsystem itself. The only reason to start or stop the timers in simulation is to inject some sort of error case you want to catch. 

This code won't run on the robot

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,221 @@
...
+    checkCurrent();
+    checkVoltage();
+    // checkMinute();
+  }
+
+  public double getVoltage() {
+    if (Robot.isSimulation()) {
+      return (int) simVolt;
+    } else {
+      return RobotController.getBatteryVoltage();
```
azvanderpas: I don't think you need a different return value for simulation. Since we are using the RoboRioSim to set the voltage it should get returned properly here.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,221 @@
...
+        .withProperties(Map.of("colorWhenFalse", "red"));
+    sbVoltage =
+        btTab.add("Current Battery Voltage", 0).withSize(2, 2).withPosition(0, 0).getEntry();
+    sbSimVoltage = btTab.add("Simulation Voltage", 0).withSize(2, 2).withPosition(2, 0).getEntry();
+    sbTimer = btTab.add("Timer", 0).withSize(2, 2).withPosition(0, 2).getEntry();
+    sbTimerHighCurrent =
+        btTab.add("High Current Timer", 0).withSize(2, 2).withPosition(0, 4).getEntry();
+    sbTimerChange = btTab.add("Change Timer", 0).withSize(2, 2).withPosition(4, 0).getEntry();
+    timer.start();
+    currentTimer.start();
```
azvanderpas: The high current time probably shouldn't start automatically

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,221 @@
...
+    } else if (timer.hasElapsed(Constants.timeInSecondsGeneralRed)) {
+      DriverStation.reportWarning("Change the Battery!", false);
+      // playAudio(Constants.audiofilepath);
+      return true;
+    } else if (currentTimer.hasElapsed(Constants.timeInSecondsHighCurrentYellow)) {
+      DriverStation.reportWarning("Change the Battery!", false);
+      // playAudio(Constants.audiofilepath);
+      return false;
+    } else if (timer.hasElapsed(Constants.timeInSecondsGeneralYellow)) {
+      DriverStation.reportWarning("Change the Battery!", false);
```
azvanderpas: Red and yellow messages are no longer distinguishable

src/main/java/frc/robot/subsystems/BatterySimulation.java
```diff
@@ -0,0 +1,65 @@
...
+  private double voltageToSubtract = 0.0001;
+  private BatterySubsystem batterySubsystem;
+
+  public BatterySimulation(BatterySubsystem batterySubsystem) {
+    this.batterySubsystem = batterySubsystem;
+    RoboRioSim.setVInVoltage(14.0);
+  }
+
+  @Override
+  public void simulationPeriodic() {
```
azvanderpas, line 23: The different tests should be sequenced in a better way. Since the whole method runs each 20 ms loop there are probably some conflicting conditions happening. Using JUnit testing framework may make some of this easier, but using getFPGATimestamp() to only run tests over a specific interval of runtime would also work. Something like this:
```
double now = Timer.getFpgaTimestamp();
if (now >= 0 && now < 1.0) {
    // Do voltage check test
} else if (now >=1.0 and now < 30.0) {
    // Do timer test
}
// else if (etc....)
```

azvanderpas (dismissed):

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: Fixed, current code for some reason didn't push properly

src/main/java/frc/robot/subsystems/BatterySimulation.java
calvinwt: Fixed, changed it to change sim current

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done, removed the variable

calvinwt: changed it to return with a RoboRioSim.getVInVoltage() method, should I just use the getBatteryVoltage() one instead?

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: fixed

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: fixed, messages are now distinguishable

src/main/java/frc/robot/subsystems/BatterySimulation.java
calvinwt: fixed, timing is now better

src/main/java/frc/robot/subsystems/BatterySimulation.java
```diff
@@ -27,39 +20,34 @@ public BatterySimulation(BatterySubsystem batterySubsystem) {
...
+      if (RoboRioSim.getVInVoltage() < Constants.minVoltageRedDouble) {
+        System.out.println("Got to assert voltage");
+        assertEquals(false, batterySubsystem.checkRedVoltage());
+      }
+    } else if (now >= 5.0 && now < 25.0) {
+      if (batterySubsystem.getGeneralTimer() > Constants.timeInSecondsGeneralRed) {
+        System.out.println("Got to assert general timer");
+        assertEquals(true, batterySubsystem.checkTimer());
+      }
+    } else if (now >= 25.0 && now < 50.0) {
```
azvanderpas: Tie time durations to your constants. Your change to timeInSecondsGeneralRed breaks this test now.

src/main/java/frc/robot/subsystems/BatterySimulation.java
```diff
@@ -27,39 +20,34 @@ public BatterySimulation(BatterySubsystem batterySubsystem) {
...
-    //Enable the robot after 10 seconds to rest high current timer
-    if (batterySubsystem.getGeneralTimer() > 10.0 ) {
+    // Enable the robot after 10 seconds to test high current timer
+    if (batterySubsystem.getGeneralTimer() > 10.0) {
       DriverStationSim.setEnabled(true);
     }
 
-    //Forcefully sets the voltage when enabling the robot.
+    // Forcefully sets current when robot is enabled to simulate use
     if (DriverStationSim.getAutonomous() == true || DriverStationSim.getEnabled() == true) {
```
azvanderpas: What is the autonomous check for?

src/main/java/frc/robot/subsystems/BatterySubsystem.java
azvanderpas: Yeah, I think just using RobotController.getBatteryVoltage() works in simulation and in real

src/main/java/frc/robot/Constants.java
```diff
@@ -17,14 +17,16 @@
...
+  public static final int timeInSecondsSimTest = 5;
   public static final int timeInSecondsGeneralYellow = 360;
-  public static final double timeInSecondsGeneralRed = 60.0; //Changed for testing
+  public static final double timeInSecondsGeneralRed = 20.0; // Changed for testing
   public static final int timeInSecondsHighCurrentYellow = 360;
-  public static final double timeInSecondsHighCurrentRed = 20.0; 
+  public static final double timeInSecondsHighCurrentRed = 20.0;
+  public static final double simCurrentLow = 10.0;
+  public static final double simCurrentHigh = 13.0;
   public static final double maxBatteryCurrent = 12.3; // TODO: Put a real value here
```
azvanderpas: I'd suggest a better name. Max current implies it isn't safe to operate above this current. 
How about highCurrentBatteryThreshold?

src/main/java/frc/robot/subsystems/BatterySimulation.java
```diff
@@ -27,39 +20,34 @@ public BatterySimulation(BatterySubsystem batterySubsystem) {
...
-      batterySubsystem.startHCTimer();
-      //RoboRioSim.setVInVoltage(12.1);
+      RoboRioSim.setVInCurrent(Constants.simCurrentHigh);
     } else {
-      //RoboRioSim.setVInVoltage(14.0);
-      batterySubsystem.stopHCTimer();
+      RoboRioSim.setVInCurrent(Constants.simCurrentLow);
```
azvanderpas: The constants here seem a little unnecessary since you already have a constant for what is high current. You might do something like RoboRioSim.setVInCurrent(Constants.maxBatteryCurrent+1) for high current and -1 for low.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,223 @@
...
+    } else {
+    }
+  }
+
+  public boolean checkRedVoltage() {
+    if (getVoltage() > Constants.minVoltageRed) {
+      return true;
+    } else {
+      return false;
+    }
```
azvanderpas:
```suggestion
    return getVoltage() > Constants.minVoltageRed;

```

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,223 @@
...
+  private boolean colorOn; // Variable for color change testing
+  // private int minutesPassed;
+
+  public BatterySubsystem() {
+    // init();
+    this.resetTimers();
+    btTab = Shuffleboard.getTab("Battery");
+    // Change 16 to desired value
+    Shuffleboard.getTab("Battery")
+        .addBoolean("Battery Voltage", () -> getVoltage() > Constants.minVoltageRed)
```
azvanderpas:
```suggestion
        .addBoolean("Battery Voltage", () -> checkRedVoltage())
```

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,223 @@
...
+    return currentTimer.get();
+  }
+
+  public void startHCTimer() {
+    currentTimer.start();
+  }
+
+  public void stopHCTimer() {
+    currentTimer.stop();
+  }
```
azvanderpas: I don't think we want to expose the timer controls as public. Either make these private or better just use currentTimer directly within this class and delete these.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,223 @@
...
+    if (check > 0 & check < 0.025 & getTimer() > 1.0) {
+      minutesPassed += 1;
+    } else {
+    }
+  }
+  */
+
+  public int getMinute() {
+    return (int) Math.floor(timer.get() / 60.0);
+  }
```
azvanderpas: Clean up unused methods and commented out code

src/main/java/frc/robot/subsystems/BatterySimulation.java
calvinwt: Fixed, tied to constants and reorganized for better flow

src/main/java/frc/robot/subsystems/BatterySimulation.java
calvinwt: autonomous check removed, it was unnecessary

src/main/java/frc/robot/Constants.java
calvinwt: done, renamed to highCurrentBatteryThreshold

src/main/java/frc/robot/subsystems/BatterySimulation.java
calvinwt: done

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done, timer values are now saved to the robot with Preferences

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -153,6 +167,10 @@ private void initCommands() {
       driveBaseTeleopCommand = new DriveBaseTeleopCommand(driveBaseSubsystem);
       driveBaseSubsystem.setDefaultCommand(driveBaseTeleopCommand);
     }
+    // if (batterySubsystem != null) {
+    //  batteryCommand = new BatteryCommand(batterySubsystem);
+    //  batterySubsystem.setDefaultCommand(batteryCommand);
+    // }
```
azvanderpas: Doesn't look like BatteryCommand is used and everything is done in periodic. Remove the class and references to it.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,199 @@
...
+    return currentTimer.get() + storedHighCurrentTime;
+  }
+
+  private void startHCTimer() {
+    currentTimer.start();
+  }
+
+  private void stopHCTimer() {
+    currentTimer.stop();
+  }
```
azvanderpas: Methods usually only need to be created to add a capability/access to the object from an outside source (public methods), or to reuse code that is commonly used and relatively long. Since this case is neither, I would rather just use currentTimer.start and stop directly where we needed it instead of having methods for them.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,199 @@
...
+      DriverStation.reportWarning("Change the Battery Soon! (VY)", false);
+    } else {
+    }
+  }
+
+  public boolean checkRedVoltage() {
+    return getVoltage() > Constants.minVoltageRed;
+  }
+
+  public void resetTimers() {
```
azvanderpas, line 175: Good work figuring the Preferences out!
One addition I would make is to save the last 'low current' voltage as a Preference. 
Then if you read a higher voltage on startup, reset the timers because the battery has likely been changed.

src/main/java/frc/robot/subsystems/BatterySubsystem.java
```diff
@@ -0,0 +1,199 @@
...
+      storedGeneralTime = Preferences.getDouble("Battery General Timer", 0.0);
+      storedHighCurrentTime = Preferences.getDouble("Battery High Current Timer", 0.0);
+    }
+    timer.reset();
+    currentTimer.reset();
+  }
+
+  // TODO: Find a way to make these not reset every redeploy. Possibly make them save their values
+  // to a file, then
+  // retrieve that file when the robot starts.
```
azvanderpas: Looks like you've solved this already. Delete comment

azvanderpas (dismissed):
Will approve with the clean up tasks. Battery changed checking can go in a new PR

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done, but still needs testing

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: done, methods have been removed

src/main/java/frc/robot/RobotContainer.java
calvinwt: done, command has been removed

src/main/java/frc/robot/subsystems/BatterySubsystem.java
calvinwt: did some testing, seems like no matter what, if you reboot the roborio it clears the values of preferences. Either way, performance should be as intended

azvanderpas (approved):

